PR #194, opened 2015-08-17
Update README

skddc:
At the moment it is not clear what's happening and people are trying to use master regularly. Let's add a warning to the README at least.


silverbucket:
agreed, thanks!

PR #202, opened 2017-01-26
Restore LICENSE

abliss:
I made this commit a merge, whose right-hand side was the last commit that had a LICENSE, so that it would be easy to check that the file has not changed since then. It does make the graphical log look a little funny though.


abliss:
Ack, I guess github doesn't offer any way to choose which parent you view the diff from? If you pull this locally and use `git show` you can check that I didn't mess up the license file.

silverbucket:
Thanks @abliss !

PR #204, opened 2017-02-01
Add CORS headers to static assets

skddc:
This is needed for dynamically loading them via Fetch, e.g. in ServiceWorker scripts.

It also serves `socket.io.js` from the same directory root as the other assets, to make it less confusing.


silverbucket:
@skddc awesome! thanks

silverbucket:
Oh, this breaks the examples, will fix those and then push a new release.

PR #207, opened 2017-02-05
Missing an ending quotation mark in the dummy example.

djenniex:
This is causing issues running the dummy example.


skddc:
Thank you!

PR #210, opened 2017-03-19
Require a halfway modern node version

### LINKED ISSUE #206: sockethub won't start on master with node v4.2.6
```
abliss@localhost:~/proj/sockethub$ bin/sockethub


UNCAUGHT EXCEPTION - something horrible has happened!

Error: SyntaxError: Unexpected token [
    at module.exports (/home/abliss/proj/sockethub/lib/bootstrap/platforms.js:52:17)
    at Object.<anonymous> (/home/abliss/proj/sockethub/lib/bootstrap/init.js:10:53)
    at Module._compile (module.js:410:26)
    at Object.Module._extensions..js (module.js:417:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Module.require (module.js:354:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/home/abliss/proj/sockethub/lib/sockethub.js:16:18)
    at Module._compile (module.js:410:26)
    at Object.Module._extensions..js (module.js:417:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Module.require (module.js:354:17)
```

skddc:
There are already features used in the code, which aren't available by default on versions older than 5. LTS is 6.x and current is 7.7.2, so I think this is a reasonable requirement.

closes #206

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,5 +59,8 @@
   "scripts": {
     "test": "node_modules/.bin/jaribu",
     "start": "DEBUG=* ./bin/sockethub"
+  },
+  "engines": {
+    "node": ">= 5.0.0"
   }
 }


skddc:
Btw, if for some reason Sockethub should still officially support older versions, then I think those should be added to https://github.com/sockethub/sockethub/blob/master/.travis.yml, so that at least it's not possible to accidentally break them in the future.

silverbucket:
@skddc thank you! yeah I've started using some more modern API stuff so it's really a requirement now.

PR #218, opened 2017-05-13
[WIP] feature/client replay

silverbucket:
As discussed in https://github.com/67P/hyperchannel/issues/77 we need a way to recover basic state if the client looses connection with the server. In most cases, this won't be due to the Sockethub server crashing, but rather intermitted/unreliable internet connectivity, however this handles things assuming there's no longer any state active on the Sockethub service:

Outgoing messages are checked for a few types of things and stored in Map()'s
  * credentials
  * activity-objects
  * messages that are `join` or `leave`

When a `reconnect` event is received, these 3 types of objects are re-sent to Sockethub to ensure it has 1. credentials for each account registered so far 2. any activity-objects that were declared in the previous session 3. any channels that were joined before are joined again.

If Sockethub hadn't crashed and still has all of that existing state, there shouldn't be any effect on it's state, so no harm no foul in that case. 

Wondering your thoughts @galfert and @skddc on this. Is this the right way to handle disconnects?


silverbucket:
ping @galfert @skddc

lib/client.js
```diff
@@ -1,29 +1,22 @@
 (function (global, factory, undefined) {
 
-  if ( typeof module === 'object' && typeof module.exports === 'object' ) {
-    var ActivityStreams = require('activity-streams');
+  if ( typeof module === "object" && typeof module.exports === "object" ) {
```
raucao: Is there a reason to change to double quotes everywhere?

lib/client.js
```diff
@@ -60,30 +42,101 @@
...
+          case "activity-object":
+            if ((content.actor) && (content.actor["@id"])) {
+              storedActivityObjects.set(content.actor["@id"], content);
+            }
+            break;
+          case "message":
+            if (content["@type"] === "join") {
+              var actor = content.actor["@id"] || content.actor;
+              var target = content.target["@id"] || content.target;
+              var key = actor + "-" + target;
```
raucao: These 3 lines are exactly the same as below, so couldn't they be outside of the `if`?

lib/client.js
```diff
@@ -60,30 +42,101 @@
...
+            } else if (content["@type"] === "leave") {
+              var actor = content.actor["@id"] || content.actor;
+              var target = content.target["@id"] || content.target;
+              var key = actor + "-" + target;
+              storedJoins.delete(key);
+            }
+            break;
+        }
+        self.socket._emit(channel, content);
+      } else {
```
raucao: It's possible to not wrap all code in here in another `if` by moving this to the top as `if (typeof channel !== "string") {` and then continuing without `else`.

silverbucket:
@skddc thanks for the review, will update the code later. As for your general comments. The testing steps that I was using was loading up sockethub+hyperchannel and connecting to channels, etc. then Cntrl-C'ing sockethub. Then after a few seconds starting sockethub up again. The reconnect event will fire, the credentials, activity objects and channels objects will be replayed to sockethub from the client library, and you should be re-joined to the channels. 

Not sure how to approach writing any tests for that,  Though some client-library-only tests could be written I suppose which mock the socket.io object and throw events, waiting for specific objects to be replayed which were sent beforehand. 

I'm still not sure if this is the right approach to solving the issue though which is the primary thing to consider and would love more feedback on it from you and @galfert. If so, I can add tests and improve the code, etc.

skddc:
Ok, thanks. Will try it out and try to form an opinion on the general idea.

galfert:
I will try to have a look at this later this evening.

galfert:
@silverbucket how did you test this? Both in Chrome and Firefox, when I go into offline mode, everything still stays connected. I can send and receive messages and join or leave channels.

skddc:
@galfert He explained it:

> The testing steps that I was using was loading up sockethub+hyperchannel and connecting to channels, etc. then Cntrl-C'ing sockethub.

galfert:
Oh, thanks. I missed that.

skddc:
> Both in Chrome and Firefox, when I go into offline mode, everything still stays connected. I can send and receive messages and join or leave channels.

Are you using SH from localhost maybe?

galfert:
Ok, that worked. All channels were re-joined after reconnecting, including those that I tried to join while offline.

When trying to send a message while offline (to a channel I already had joined), I did not get any failure feedback. But after reconnecting, I got a SH failure event with error "cannot send message to a channel of which you've not first `join`ed."

skddc:
> Ok, that worked. All channels were re-joined after reconnecting, including those that I tried to join while offline.

w00t!

> When trying to send a message while offline (to a channel I already had joined), I did not get any failure feedback. But after reconnecting, I got a SH failure event with error "cannot send message to a channel of which you've not first joined."

Sending new messages is not actually preserving state, so I think that shouldn't even be tried automatically in the first place (and needs instant feedback locally). As an app developer I want to know it failed, cache the message and then either retry later, or have the user delete it before then if they want.

silverbucket:
@skddc @galfert - so, things worked as expected? any feedback on behaviour you'd like added, removed or changed? I take it that it might be a good idea to have the client reject any messages from the app automatically? (send them back on the failure channel) for example, with messages being sent from the app to sockethub, while the socket.io `reconnected` event hasn't yet fired.

galfert:
> I take it that it might be a good idea to have the client reject any messages from the app automatically?

Yes, some kind of feedback would be good.

lib/client.js
silverbucket: Crockford made a case against using single quotes as the default quotation mark. Can't find a link at the moment, but more an more I'm running into eslint or tslint which complains about single quotes, so I'm trying to force myself into the habit.

galfert: I recently read somewhere (can't remember exactly where though), that most people moved away from Crockford style to [Standard JS](https://github.com/feross/standard). And that one recommends single quotes.

raucao: Namedropping isn't exactly an argument for either side. I think if anything, it should be consistent, and if we want to change it to double quotes (which I personally don't see a reason for, but I couldn't care less), then we should do it across the board and introduce linting for it, instead of just randomly starting to mix in a different style.

silverbucket: I'm not "namedropping" he made a good point that I reluctantly had to agree with at the time, however I wasn't able to find a link. It has to do with simplifying the language to it's essential parts and not using parts of the language which serve no purpose or can cause issues. Something like "What does a simplified, non-redundant, JavaScript language look like". And choosing double quotes over single-quotes was one of them. I'm working on getting linting set up across the projects but I haven't been able to do so yet.

silverbucket: @galfert looks interesting will give it a read. Though I'm not sure how we can qualify the term "most people" :)

silverbucket:
OK @galfert and @skddc - here's the latest update, I notify the client when a message has been sent while in an offline/disconnected state (on the failure channel). and replay successful (completed during a connected/reconnected state) join, activity-object, and credential messages when we return to a reconnect state. Now with tests. Feedback welcome!

galfert:
Cool, I will give it a try today.

lib/client.js
silverbucket: @galfert oh no, no semicolons? spaces after named functions? D:

galfert:
I tested this by <kbd>Ctrl+C</kbd>ing Sockethub.

I still did not see any messages on the failure channel when trying to send messages while offline.

After I started the Sockethub server again, the first message I received was:

```json
{
    "@type": "announce",
    "actor": {
        "@id": "irc://irc.freenode.net",
        "@type": "service"
    },
    "context": "irc",
    "object": {
        "@type": "content",
        "content": {
            "channel": {
                "idlength": {},
                "kicklength": 0,
                "length": 50,
                "limit": {
                    "#": 120
                },
                "modes": 4,
                "topiclength": 390,
                "types": "#"
            },
            "modes": {
                "channel": "CFILMPQSbcefgijklmnopqrstvz",
                "maxlist": {
                    "bqeI": 100
                },
                "param": "bkloveqjfI",
                "prefixes": "@+",
                "prefixmodes": {
                    "o": "@",
                    "v": "+"
                },
                "types": {
                    "a": "eIbq",
                    "b": "kov",
                    "c": "flj",
                    "d": "CFLMPQScgimnprstz"
                },
                "user": "DOQRSZaghilopswz"
            },
            "network": {
                "hostname": "leguin.freenode.net",
                "ircd": "ircd-seven-1.1.4",
                "maxtargets": {
                    "ACCEPT": 0,
                    "KICK": 1,
                    "LIST": 1,
                    "MONITOR": 0,
                    "NAMES": 1,
                    "NOTICE": 4,
                    "PRIVMSG": 4,
                    "WHOIS": 1
                },
                "name": "freenode",
                "nicklength": 16
            }
        }
    },
    "published": "2017-06-15T11:01:36.017Z",
    "target": {
        "@id": "irc://irc://hyper_galfert@irc.freenode.net@irc.freenode.net",
        "@type": "person"
    }
}
```

Notice the weird target ID ("irc://irc://hyper_galfert@irc.freenode.net@irc.freenode.net").

The second message was on the failure channel:

```json
{
    "@type": "send",
    "actor": {
        "@id": "irc://hyper_galfert@irc.freenode.net",
        "displayName": "hyper_galfert"
    },
    "context": "irc",
    "error": "cannot send message to a channel of which you've not first `join`ed.",
    "object": {
        "@type": "message",
        "content": "testing from offline"
    },
    "target": {
        "@id": "irc://irc.freenode.net/#hyperbla",
        "@type": "room",
        "displayName": "#hyperbla"
    }
}
```

silverbucket:
@galfert thanks for testing this. I'm trying to figure out the best way to reproduce this, could you post the JSON objects you're sending when offline that get replayed back incorrectly?

galfert:
The message object I was trying to send:

```json
{
    "@type": "send",
    "actor": "irc://hyper_galfert@irc.freenode.net",
    "context": "irc",
    "object": {
        "@type": "message",
        "content": "hello from offline"
    },
    "target": {
        "@id": "irc://irc.freenode.net/#hyperbla",
        "@type": "room",
        "displayName": "#hyperbla"
    }
}
```

You can also easily test it yourself, using the `feature/handle-join-failures` branch of Hyperchannel.

As far as I can see it, Sockethub tries to send the message before it (re-)joined the channel. After the failure message I posted earlier, I can see completed events for the several channel joins.

PR #219, opened 2017-06-09
eslintrc enforcing some code confirmity

silverbucket:
So I finally managed to put together a starting .eslintrc file. Review very much welcome.

The biggest controversy is of course about single vs double quotes. I've always been a single quote kind of person, but Douglas Crockford made a good point in this video about continuing to simplify the language as much as possible as we go along, things like removing var in place of let and const, double quote instead of single quote, undefined instead of null, and backs up each point. Maybe hear him out and then tell me what you guys think? (@galfert @skddc )

https://www.youtube.com/watch?v=NPB34lDZj3E


silverbucket:
Oh crap, I must have branched this off the client_reply branch, so I guess it should wait to be merged until that one is merged. Not sure how I got a bunch of old commits from master in here too though. :/

silverbucket:
Thoroughly confused by the commit history... anyone have any ideas?

skddc:
You can just create a new branch from master and cherry-pick the couple of commits reg. eslint to it.

skddc:
Reg. Crockford: could you please either link the time when he talks about why double quotes make more sense than single quotes (as you yourself like single quotes better, and the rest of us too)? I don't want to watch 45 minutes of someone explaining why his flavor of syntax should be the canonical one. ;)

silverbucket:
It's pretty early in the talk, he explains the concept of his talk then it's one of the first examples. Sorry I'm not at my computer right now
On Jun 9, 2017, at 18:29, Sebastian Kippe <notifications@github.com<mailto:notifications@github.com>> wrote:

Reg. Crockford: could you please either link the time when he talks about why double quotes make more sense than single quotes (as you yourself like single quotes better, and the rest of us too)? I don't want to watch 45 minutes of someone explaining why his flavor of syntax should be the canonical one. ;)

silverbucket:
@skddc He starts on double vs single quotes around the 9:45 mark.

galfert:
+1

skddc:
I finally watched that bit, and there's no actual argument in there. Also, the bit before that is about getting rid of unnecessary things as well, like tabs vs. spaces. At least with those, there's an argument for literally not being able to remove spaces, so the other should go. But in the end those are all just personal preferences (the world will not abide by one man's talk), and he ends the bit with a joke proving that that's the case.

So, if this talk doesn't bring up any good arguments, other than "if we removed one of the two, my personal preference would be keeping double quotes", then we might as well go with our personal preference, the single quote. Also because that's in line with JavaScript Standard Style, which is more widely used than whatever the name would be for Crockford's personal prefs (The Good Parts™ I guess?).

skddc:
@silverbucket I think you might have missed my response above, back when I wrote it. There's also a merge conflict to be resolved now.

skddc:
@silverbucket Did you see my last comment? ^^

silverbucket:
@skddc yeah, I've been meaning to go about cherry picking this branch, it's just not something I am very good at (I have a tendency to irreversibly destroy things when I start trying to mess with commit histories) but you make a fair enough point regarding single vs. double quotes. I prefer single quotes myself, I was jus trying to be as uniform as possible, we can still do that with the eslintrc, I'll just remove the double quote config option. 

I'm not sure the best way to revert double quotes to single, as it seems to be spread among other commits with other changes. I could always just re-commit with single quotes...

skddc:
> I'm not sure the best way to revert double quotes to single, as it seems to be spread among other commits with other changes. I could always just re-commit with single quotes...

Yes, I'd just change it later in a new commit, after rebasing.

skddc:
Friendly bump. ;)

PR #220, opened 2017-09-29
Remove comments from the JSON config: JSON doesn't allow comments

gregkare:
I found out about the problem while working on the Docker setup (I'll send a PR soon, using Docker Compose)

The config entry for the Redis URL is explained in the README. If you
want comments you need to use YAML or another format that has comments

### DIFF
--- a/config.json.example
+++ b/config.json.example
@@ -32,8 +32,6 @@
   "redis": {
     "host": "redis",
     "port": 6379
-    // if defined overrides host + port pair, env: REDIS_URL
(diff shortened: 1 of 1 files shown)


silverbucket:
thanks @gregkare do you need a new version released for `hyperchannel` use or are we currently using the github url?

gregkare:
@silverbucket Hey, we're deploying from GitHub for now, I'll let you know if we need a new release

PR #223, opened 2018-09-14
Feature/eslint2

silverbucket:
Alright, I'm finally getting around to merging these PRs starting with the linting, I set the single, instead of double quotes rule (while allowing `` for escaped strings) and am enforcing es6 syntax, uniform indention, and a few other pieces. We can evolve this or adjust as we see necessary down the line.


skddc:
LGTM. :+1:

silverbucket:
OK will merge, thanks!

PR #224, opened 2018-09-15
Feature/rework connection handling

silverbucket:
This is a fairly substantial PR that re-works the way sockethub handles unexpected/unrecoverable events at the worker+platform level. I started this re-work in January 2018, but it was not ready for merge because renaming users functionality hadn't been done yet (due to `uri` mapping issues), however it has been running and tested and proven to be much more stable as a whole than previous versions.

Over the past couple weeks I've worked on implementing renaming of a users, no longer relying on the users `uri` as the main index, but rather a random UID for each platform instance (generated by the worker), along with a mapping table of `uri` -> UID.


silverbucket:
This is ready for review @skddc @galfert

silverbucket:
@galfert @skddc - i know this is quite a big PR and as it's more of a reworking of underlying stuff and not exactly a feature or concise fix, it's probably not easy to review without knowing a lot about the way the code works. Do either of you want to review it or should I just merge it? I don't recall exactly what our agreement was with PR reviews for Sockethub.

skddc:
Sorry, only just now found the time to have a look at this. But as you say, it's rather difficult to review for someone not overly familiar with the code.

PR #231, opened 2018-10-09
Feature/code cleanup

silverbucket:


skddc:
Nice! These are the things that make understanding the code and then contributing so much easier.

PR #238, opened 2019-04-11
Re-add unminified activity-streams.js file

### LINKED ISSUE #237: activity-streams should provide sourcemaps
See https://github.com/67P/hyperchannel/issues/161

<strike>We are using https://sockethub.kosmos.org:10550/activity-streams.js, but now only https://sockethub.kosmos.org:10550/activity-streams.min.js is provided</strike>

We're changing Hyperchannel to use the minified version, having sourcemaps would help when developing

galfert:
Fixes sockethub/sockethub#237

Removing the file broke existing apps linking to it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -410,9 +410,9 @@
           "integrity": "sha1-7QMXwyIGT3lGbAKWa922Bas32Zg="
         },
         "camelcase": {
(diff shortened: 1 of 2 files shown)


silverbucket:
thanks @galfert sorry i didn't get this sooner!

PR #240, opened 2019-08-10
Deps update

silverbucket:


skddc:
You don't earn kredits if you don't add kredits labels... :wink: 

Also, if you do add kredits labels, make sure the PR description is written nicely, so that it's clear what the contribution was when just looking at the [kredits contribution dashboard](https://kredits.kosmos.org/).

PR #241, opened 2019-08-10
Migrate to TypeScript

silverbucket:
* Switch to using TypeScript for development. 
* Refactored the bootstrap and config process to be more flexible.
* Switch to es6 imports/exports for everything except `bootstrap/platforms.js` due to the dynamic loading of platforms (plugins).


silverbucket:
Need to work still on refactoring all of the unit tests, but have the core done.

PR #245, opened 2019-12-19
Feature: Switch to lerna

silverbucket:


skddc:
Looks like something went wrong with Git here. 345 commits difference to master can't be right for this feature...

silverbucket:
@skddc It imports the git history for each of the repositories imported with lerna.

skddc:
Oh wow, that's actually pretty cool!

skddc:
Would be extremely useful to get this finished sometime this week. We found some critical issues with Sockethub/XMPP (GitHub issues incoming), but nobody wants to risk having their potential PRs conflict with this major re-org.

@silverbucket Let us know if/how we can help.

skddc:
> @silverbucket Let us know if/how we can help.

^^

silverbucket:
You, it's still on my radar, I've just been really busy this month and I haven't had extra time. This branch is probably not going to be the one that's merged, I've started another but having done much with it yet. If there's someone that knows or wants to learn more about lerna it wouldn't hurt to have someone to discuss the versioning decisions that need to be made with it and help define a release process. But not necessary if no one has time, and I'll have more opportunities to get this done after January.

skddc:
> it wouldn't hurt to have someone to discuss the versioning decisions that need to be made with it and help define a release process

I don't see the connection between moving to Lerna and changing the release process. It's only the Git repos that are being merged, while the npm modules can stay the same until that's all done. The npm release process can be exactly the same as before for now. Or am I missing something?

galfert:
I think it's because Lerna has those two modes for versioning and you have to choose one when creating the config: https://github.com/lerna/lerna#how-it-works.

If it was just for Sockethub and the platforms, I would tend towards a common version for all packages. It makes it more clear that those packages should all work with each other when they all have the same version. But I'm not so sure about other libraries, like `activity-streams.js`. Increasing the major version number of that, just because there is a breaking change in one of the Sockethub platforms doesn't make much sense to me.

PR #254, opened 2020-02-18
Replace cdd call with cd -

galfert:
In the `dev` script in package.json I noticed a call to `cdd` at the end. I didn't recognize that commend and found that I don't have such a command (on macOS).
I don't know if it's a custom script or something Linux specific, but I assume it's for changing back to the original directory after the script previously changed into `packages/sockethub`. I replaced it witch `cd -`, which does exactly that, but AFAIK is in the POSIX standard. So it should be available in most shells.

I also added it to the end of the `start` script as well, because that also changes directories.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,8 +12,8 @@
     "test": "lerna run test",
     "travis": "lerna run travis",
     "build": "lerna run build",
-    "dev": "[[ -f config.json ]] && cp config.json packages/sockethub/; cd packages/sockethub && yarn run dev; cd -",
-    "start": "[[ -f config.json ]] && cp config.json packages/sockethub/; cd packages/sockethub && yarn run start; cd -"
+    "dev": "[[ -f config.json ]] && cp config.json packages/sockethub/; cd packages/sockethub && yarn run dev",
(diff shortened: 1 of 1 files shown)


skddc:
May I suggest to use `pushd` and `popd` for that? I think it's much safer across environments and in scripts like that.

gregkare:
It turns out the `cd -` command is not needed, you can remove it and you are kept at the root of the repo. Only the subshell is moving to the `packages/sockethub` dir

galfert:
Oh cool, didn't know that. Will remove it then.

silverbucket:
Awesome, thanks. BTW can't figure out how to mark the PR as "approved" by me as I'm requested as a reviewer. Anyway, will merge and publish a new release.

silverbucket:
Nevermind about the publishing bit, :D forgot this was just for the repository and not for any of the packages.

galfert:
> BTW can't figure out how to mark the PR as "approved" by me as I'm requested as a reviewer.

In the "Files changed" tab there is a green button "Review changes" at the top right.

PR #275, opened 2020-07-28
Remove all platform instances when main process exits

### LINKED ISSUE #271: Thread management when parent dies ensure children are terminated
When the main sockethub thread throws an exception and dies, the children will still be running. Not sure if there's a way to ensure the children die, but needs to be investigated a bit more.

galfert:
Closes #271

When the main process exits, either by receiving a TERM or INT signal, or when an unhandled exception occurs, all platform instances are removed. So no child processes are left behind.

### DIFF
--- a/packages/sockethub/bin/sockethub
+++ b/packages/sockethub/bin/sockethub
@@ -2,26 +2,24 @@
 const baseDir = process.env.DEV ? 'src' : 'dist';
 const SharedResources = require(`./../${baseDir}/shared-resources`);
 
-process.on('uncaughtException', (err) => {
+process.once('uncaughtException', function (err) {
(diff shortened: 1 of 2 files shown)


silverbucket:
@galfert awesome! I'll check this out and try to break it

silverbucket:
@galfert With a normal control-c or exception thrown, the child process is killed, however it seems that if I kill the main thread using `kill -9` the children thread still lives. Perhaps there's no way around that issue.

skddc:
> f I kill the main thread using `kill -9` the children thread still lives

That is to be expected. You would need to adjust your kill command, or use `pkill -P` for example. In any case, I think it's not the responsibility of a program to guard against `kill -9` operations.

silverbucket:
LGTM thanks @galfert !

Issue #270, opened 2020-07-23
Handle non-instance-specific platform threads


silverbucket:
non-persistent threads (eg. feeds) need to verify they are started up correctly and managed efficiently. Some work still needs to be done here.

PR #278, opened 2020-09-08
Support for static platform threads (eg. global/shared) 

silverbucket:
Resolves issue #270, handling platform threads which should be up and running (a single instance) and shared amongst users.


skddc:
Just a note about the title: consider it making sense for someone looking at the kredits dashboard (just a little bit).

silverbucket:
Fixed some issues with unit tests, this PR is now ready for review. Updated the title as @skddc suggested

skddc:
> this PR is now ready for review

Great! Could you give potential reviewers a hint for an easy way to test the fixes?

packages/sockethub/src/resource-manager.ts
```diff
@@ -31,7 +31,9 @@ function resourceManagerCycle() {
         }
       }
 
-      if (platformInstance.sessions.size <= 0) {
+      if (platformInstance.global) {
+        // static platform for global use, don't do resource management
```
galfert, line 35: Wouldn't it be better to have a `continue` statement here instead of doing nothing?

silverbucket:
This ticket is for platform threads which don't contain any user specific data. There's no login or active session that is specific to the user who initiated the request. Therefore, the in-out data to-from the platform doesn't need to be sandboxed to the user, and a single thread can be shared among anyone who wants to use it. The feeds platform is an example of this. Making multiple requests to the feeds platform from separate sessions shouldn't result in 1. multiple feeds instances 2. the feed instance being cleaned up/killed when no more sessions are referencing it. 

Unfortunately, the feeds platform itself isn't actually working, likely due to library API changes. However the request is sent in and out from the feeds platform so the testing of the functionality should still be OK.

galfert:
I didn't know how to test the feeds. But I can confirm that this PR doesn't break anything when using Hyperchannel :)

packages/sockethub/src/resource-manager.ts
```diff
@@ -33,6 +33,7 @@ function resourceManagerCycle() {
 
       if (platformInstance.global) {
         // static platform for global use, don't do resource management
+        return;
```
galfert, line 36: @silverbucket Why did you use `return` instead of `continue` now? That will stop the whole `for` loop that iterates over all platform instances. Is that really intended?

silverbucket: @galfert Thanks for pointing this out, I was a little distracted. It's fixed in master

PR #285, opened 2020-11-17
Changing actor name updates mappings and credentials store

### LINKED ISSUE #269: Re-implement updateCredentials
Handle platform.js updateCredential messages (when username or password are changed from within a sockethub session, those new credentials need to be updated for that sessions verification process.

silverbucket:
* updateCredentials -> updateActor
* Mappings and store are updated with new lookup identifier (hash of actor string)
* Destruction of platformInstance consolidated and simplified
* Additional unit tests and coverage

Resolves #269


raucao:
Cool. The Travis build is still failing. I'll review when it's green, so I can just go by the tests.

silverbucket:
@raucao seems to be failing to get a `jest` command which was listed as a peer dependency, just moved it to a dev dependency so hopefully that fixes it.

silverbucket:
@raucao seems like things are passing and green now

silverbucket:
@galfert or @raucao this PR is ready for review, ignore the code climate issue (function is 26 lines instead of 25)

packages/sockethub/src/platform-instance.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    pi.process = {
+      on: jest.fn().mockName('pi.process.on'),
+      removeListener: jest.fn().mockName('pi.process.removeListener'),
+      removeAllListeners: jest.fn().mockName('pi.process.removeAllListeners'),
+      unref: jest.fn().mockName('pi.process.unref'),
+      kill: jest.fn().mockName('pi.process.kill'),
+    };
+  });
+
+  it("should provided have properties set and be global", () => {
```
raucao: Typo?

packages/sockethub/src/platform-instance.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    expect(SharedResources.platformInstances.set).toBeCalledWith('foo bar', pi);
+  });
+
+  it('should send message to client using socket', () => {
+    pi.sendToClient('my session id', {foo: 'this is a message object'});
+    expect(SharedResources.sessionConnections.get).toBeCalledWith('my session id');
+  });
+});
+
+describe('test private instance per-actor', () => {
```
raucao: It's self-evident that this is a test, so the description should only be the thing IMO.

```suggestion
describe('Private instance per-actor', () => {
```

raucao (commented):
Left some comments. Love that there are more tests now!

One thing I would consider when using spec syntax like "describe", is to also use spec language instead of "should". For example "it should generate a failure report" would turn into "it generates a failure report", which both removes the superfluous extra word, as well as reads more like a specification than a wish list. Just my personal opinion though.

Btw, I think this PR is more of a `kredits-2`. The 3 is really a bit of a nuclear option for when it's most definitely more than 3 mediums, and usually even more than that. I had plenty of mediums that I spent more than a day working on for example. The incentive should always be to make PRs smaller (and thus easier to review and reason about) instead of slapping a 3 on ones that contain side quests. :)

silverbucket:
> Left some comments. Love that there are more tests now!
> 
> One thing I would consider when using spec syntax like "describe", is to also use spec language instead of "should". For example "it should generate a failure report" would turn into "it generates a failure report", which both removes the superfluous extra word, as well as reads more like a specification than a wish list. Just my personal opinion though.

Yeah, good suggestion. I'll do that.

> Btw, I think this PR is more of a `kredits-2`. The 3 is really a bit of a nuclear option for when it's most definitely more than 3 mediums, and usually even more than that. I had plenty of mediums that I spent more than a day working on for example. The incentive should always be to make PRs smaller (and thus easier to review and reason about) instead of slapping a 3 on ones that contain too many side quests. :)

OK, I am still pretty unsure of when to use which, so I'll try to remember that rule of thumb.

packages/sockethub/src/platform-instance.test.ts
```diff
@@ -0,0 +1,111 @@
...
+  it('is able to generate a failure reports', () => {
+    pi.registerSession('my session id');
+    expect(pi.sessions.has('my session id')).toBe(true);
+    pi.reportFailure('my session id', 'an error message');
+    pi.sendToClient = jest.fn();
+    pi.destroy = jest.fn();
+    expect(pi.sessions.size).toBe(0);
+  });
+
+  it("cleans up it's references when destroyed", () => {
```
raucao:
```suggestion
  it("cleans up its references when destroyed", () => {
```

packages/sockethub/src/platform-instance.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    pi.destroy();
+    expect(SharedResources.platformInstances.delete).toBeCalledWith('platform identifier');
+  });
+
+  it("updates it's identifier when changed", () => {
```
raucao:
```suggestion
  it("updates its identifier when changed", () => {
```

packages/sockethub/src/platform-instance.test.ts
```diff
@@ -0,0 +1,111 @@
...
+  });
+
+  it('adds a close and message handler when a session is registered', () => {
+    pi.registerSession('my session id');
+    expect(handlers.callbackFunction).nthCalledWith(1, 'close', 'my session id');
+    expect(handlers.callbackFunction).nthCalledWith(2, 'message', 'my session id');
+    expect(pi.sessions.has('my session id')).toBe(true);
+  });
+
+  it('is able to generate a failure reports', () => {
```
raucao:
```suggestion
  it('is able to generate failure reports', () => {
```

raucao (commented):
Just some typos...

PR #290, opened 2020-11-18
Fix multiple listeners on the job queue

silverbucket:
This PR fixes a bug where multiple instances of a job queue handler could be registered, creating the problem where a join could be called before a connect was finished.  

Additionally, this fixes a related non-breaking bug where the bootstrap process is inadvertently fired every time a process is forked, when all that is needed is the config file be re-read. So this decouples the init and config reading paths entirely.


raucao (commented):
Code LGTM :+1:

PR #293, opened 2020-11-18
Activity.Object.get returns undefined

### LINKED ISSUE #292: activity-streams.js: Activity.Object.get({..obj}) returns undefined

silverbucket:
* created failing test to reproduce
* return whatever `id` is if we can't look it up and the `expand` boolean isn't set
* move test to same directory as source

Resolves #292


packages/activity-streams.js/browser/activity-streams.min.js.LICENSE.txt
```diff
@@ -0,0 +1,16 @@
...
+ * Developed and Maintained by:
+ *   Nick Jennings <nick@silverbucket.net>
+ *
+ * activity-streams is released under the MIT (see LICENSE).
+ *
+ * You don't have to do anything special to choose one license or the other
+ * and you don't have to notify anyone which license you are using.
+ * Please see the corresponding license file for details of these licenses.
+ * You are free to use, modify and distribute this software, but all copyright
+ * information must remain.
```
raucao, line 14: This looks like some leftover from dual licensing somewhere.

raucao (approved):
LGTM, but the PR title is rather cryptic for someone looking at the kredits dashboard, or release notes. And the commit history doesn't contain a commit that claims to fix the bug. Maybe just squash the commits into one when merging, so you can explain the fix in that single commit then.

packages/activity-streams.js/browser/activity-streams.min.js.LICENSE.txt
silverbucket: It's for browser packages which just contain the content in the browser directory

PR #297, opened 2020-11-19
Create LICENSE

silverbucket:


raucao:
Shouldn't licenses be per package?

silverbucket:
@raucao yeah, I'm not sure how that should work. I was going through the github community checklist and it was showing up as not done as there's nothing in the root. So I checked random number of lerna repos and they all have the LICENSE file at the root, and several had different license files in various package dirs. So I guess it's normal? Not sure how it applies though if it were to come down to it.

silverbucket:
@raucao updating the sockethub* repos to use MPL while sockethub-schemas, activity-streams.js, irc2as use MIT. The one in the root of the repo can be a default MPL as a catch all

raucao:
Cool.

Not that I want to start a whole licensing discussion, but is there a reason not to use AGPL instead? MPL seems odd to me.

silverbucket:
@raucao hehe, well, that's the question. I guess it has to do with distributed sources right? I actually just chose MPL because that's what hyperchannel is using, and I just am not fully up to date on how the GPL plays with included JS files, etc. in browsers. I remember thinking a lot about this back then and choosing GPL, and I believe the updated version has clauses for included JS, so maybe it's fine... why did you choose MPL for hyperchannel?

raucao:
Here's the original issue, but I failed to document the preceding afk convo: https://github.com/67P/hyperchannel/issues/7

IIRC it was mostly about the trademark usage. However, I think we should switch Hyperchannel to a normal GPL as well, because the MPL allows embedding it in something proprietary. Then again, maybe that makes sense for Sockethub, I don't know.

silverbucket:
Decided to stick with the GPL for now, adding it to the root and within sockethub and platform directories.

PR #299, opened 2020-11-20
add release drafter

silverbucket:


raucao:
Just a heads up: one quirk of release drafter is that before you've created your first release with it, it will keep updating the first draft with *all* closed PRs in the repo's history (at least the last few times I set it up). So, just wait with removing the extra entries, until you actually want to release...

silverbucket:
@raucao good to know, but is there still something I need to do? When I go to releases there's no draft there.

raucao:
If you installed the GitHub app for the repo and added the yml file, then the draft will appear eventually.

silverbucket:
@raucao hmm, still nothing. It exists in settings->integrations though. https://github.com/organizations/sockethub/settings/installations/13081122

raucao:
Maybe as them directly then: https://github.com/release-drafter/release-drafter

PR #303, opened 2020-12-11
Ensure sockethub-client logs and replays connect in order

### LINKED ISSUE #288: Ensure sockethub client replay also queues connect before issuing other commands
In some cases the forked platform script can get duplicate calls to listen to jobs, ending up in concurrent job processing, which can cause issues with race conditions between order of commands. (eg. connect, then join rooms)

silverbucket:
This is mostly to future proof the work we're doing in refactoring the xmpp platform, but should be here anyway. Previously a `join` would infer a connect. However if a connect or disconnect is called explicity we want to log those commands and replay them as we do for joins, credentials and activity objects. Resolves #288


galfert (approved):
The code looks ok to me.

Unfortunately I couldn't test it with Hyperchannel. One of the previous commits seems to have broken something. I can't connect to IRC anymore.

I tried to switch to a couple of commits to find the culprit, and I think it's commit 98d07ff. When going to the commit before that one (c9271c3), connecting worked again.

I couldn't see any exceptions in the logs, but saw these messages:

> sockethub:core   received (failed) job for non-existent socket qEzKQHdEJ_zUFF2HAAAA +33s
  sockethub:core   job #3 on socket osB-sQ7Vl70kStyiAAAB failed +26s
  sockethub:core   job #4 on socket osB-sQ7Vl70kStyiAAAB failed +2ms

silverbucket:
This looks like something that I had "fixed" (commented out in the past) that I remembered I'd done for Hyperchannel, but when I was paired with @raucao neither of us could think of any good reason for this - here https://github.com/sockethub/sockethub/commit/98d07ff7d7de886584a8a5307f95f24296eb5eb8#diff-24fe96dd468a08210f52130dd9819c2660399d24835eb332a73c21dd2756c25eL560 and here https://github.com/sockethub/sockethub/commit/98d07ff7d7de886584a8a5307f95f24296eb5eb8#diff-24fe96dd468a08210f52130dd9819c2660399d24835eb332a73c21dd2756c25eR571. So we re-enabled the secure logic...  I am still confused as to why we needed to make that change in the first place. 

Does it ring any bells for anyone? I should have added an explanation when I did it, but I think I was temporarily disabling it to make Hyperchannel work and then it was essentially forgotten.

raucao:
It does not ring a bell here. I have never edited that code before pairing with you.

silverbucket:
> It does not ring a bell here. I have never edited that code before pairing with you.

Yeah I made the code change, the bell would be in regard to an issue on the hyperchannel side that prompted the change. Oh well, guess we'll just sort it out now from scratch.

@galfert Since this happened several commits ago and isn't part of this PR, what do you think of merging this and opening a new ticket to debug the issue? Once immediate question is whether you're running Hyperchannel locally or elsewhere? What was the full log on the server side and JS console?

galfert:
Yes, of course. That's why I approved the PR :) 
Just wanted to note that I couldn't test it properly.

If I remember correctly, there were no logs about the failure in the JS console and the ones on the server I noted above. But I will try again and put everything I can find in a new issue.

PR #316, opened 2021-01-22
[xmpp] Switch to Jest

silverbucket:
Switching test framework to `jest` and getting some basic coverage in place. This should help greatly as we continue to refactor the code to use `xmpp.js`


packages/sockethub-platform-xmpp/src/utils.test.js
```diff
@@ -0,0 +1,12 @@
+const utils = require( "./utils");
+
+describe('buildXmppCredentials', () => {
+  it('returns correct jid', () => {
```
raucao: This doesn't seem to test for a JID.

silverbucket: Good point, updated the description

galfert (approved):
Looks ok to me, except for some things I already commented on in #318.

PR #320, opened 2021-02-10
disable tls connections for now, with reference to issue

### LINKED ISSUE #311: Connecting to IRC doesn't work anymore
When testing PR #303, I noticed that I couldn't connect to IRC with Hyperchannel anymore when using a dev Sockethub instance.

The only messages in the Sockethub logs that I noticed as potentially relevant were these:

> sockethub:core   received (failed) job for non-existent socket qEzKQHdEJ_zUFF2HAAAA +33s
  sockethub:core   job #3 on socket osB-sQ7Vl70kStyiAAAB failed +26s
  sockethub:core   job #4 on socket osB-sQ7Vl70kStyiAAAB failed +2ms

I didn't see any errors in the JS console in Hyperchannel.

I think the problem was introduced with commit 98d07ff.

@silverbucket has noted some ideas about the particular changes which could be the reason in https://github.com/sockethub/sockethub/pull/303#issuecomment-750561419:

> This looks like something that I had "fixed" (commented out in the past) that I remembered I'd done for Hyperchannel, but when I was paired with @raucao neither of us could think of any good reason for this -
>
> * https://github.com/sockethub/sockethub/commit/98d07ff7d7de886584a8a5307f95f24296eb5eb8#diff-24fe96dd468a08210f52130dd9819c2660399d24835eb332a73c21dd2756c25eL560
> * https://github.com/sockethub/sockethub/commit/98d07ff7d7de886584a8a5307f95f24296eb5eb8#diff-24fe96dd468a08210f52130dd9819c2660399d24835eb332a73c21dd2756c25eR571.
>
> So we re-enabled the secure logic... I am still confused as to why we needed to make that change in the first place.

silverbucket:
fixes #311

### DIFF
--- a/packages/sockethub-platform-irc/index.js
+++ b/packages/sockethub-platform-irc/index.js
@@ -563,6 +563,7 @@ IRC.prototype.__connect = function (key, credentials, cb) {
   // if (is_secure) {
   //   socket = new TlsSocket(netSocket, { rejectUnauthorized: false });
   // }
+  const is_secure = false;
 
   const module_creds = {
     socket: netSocket, // socket,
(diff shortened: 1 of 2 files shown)


raucao:
Why remove the port setting if it's only about TLS?

silverbucket:
> Why remove the port setting if it's only about TLS?

Because 6667 id the non-TLS port for freenode. I was thinking of the case where someone sets 'secure: false' but leaves port '7000' which wouldn't work, but it's true that not all irc servers are going to use the same clear port. Will re-add this.

PR #322, opened 2021-02-11
Use irc-socket Tls fix from silverbucket/irc-socket

silverbucket:
This PR uses the Tls irc-socket fix in https://github.com/silverbucket/irc-socket/tree/v4-alpha 
Tested on my local machine and it works at least with both secure and non-secure connection sand send/receiving messages.

### DIFF
--- a/packages/sockethub-platform-irc/index.js
+++ b/packages/sockethub-platform-irc/index.js
@@ -564,9 +564,11 @@ IRC.prototype.__connect = function (key, credentials, cb) {
     server: credentials.object.server || 'irc.freenode.net',
     realname: credentials.actor.displayName || credentials.object.nick,
     port: (credentials.object.port) ? parseInt(credentials.object.port, 10) : (is_secure) ? 6697 : 6667,
-    debug: console.log,
-    connectOptions: { rejectUnauthorized: false }
+    debug: console.log
   };
+  if (is_secure) {
+    module_creds.connectOptions = { rejectUnauthorized: false };
+  }
   this.debug('attempting to connect to ' + module_creds.server + ':' + module_creds.port);
 
   const client = IrcSocket(module_creds, is_secure ? tls : net);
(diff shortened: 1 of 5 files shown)


raucao:
I just tried to validate this fix. But both on the master branch, as well as this one, I get CORS errors from trying to connect to sockethub in the first place:

Screenshot from 2021-02-15 16-30-21

galfert (approved):
I can confirm that this fix works for me using Hyperchannel master.

raucao:
Same in both Firefox and Chromium for me. Sockethub client doesn't try to connect via Websockets at all, and fails to connect due to missing CORS headers when falling back HTTP polling.

galfert:
I just ran `yarn install` and then `DEBUG=sockethub* yarn run dev` from the project root.

I also see large diffs for yarn lock files in the root and in `packages/sockethub-platform-irc/`.

But everything works for me with both Chrome and Firefox.

When I quit the Sockethub server while Hyperchannel is still running, I can see the same error messages as you. So my guess would be that Sockethub didn't start properly for you.

raucao:
> So my guess would be that Sockethub didn't start properly for you.

Looks alright to me:

Screenshot from 2021-02-15 18-06-10

galfert:
Yes, that looks pretty much the same as here.

Can you open http://localhost:10550/ in the browser?

raucao:
> Can you open http://localhost:10550/ in the browser?

Aha: those examples work just fine, using Websockets immediately! Anything not on he same origin doesn't seem to attempt Websockets and fails at HTTP polling with the CORS errors.

silverbucket:
@raucao Hmm.. Are both Hyperchannel and Sockethub running on localhost? Becuase if I do that they work fine, but I do get weirdness if they are running on different "hosts" eg `local.dev` and `localhost` or some variant like that. `localhost` + CORS is treated with suspicion by the browser in some cases that wouldn't ordinarily be the case on two different hosts neither of which are localhost..

silverbucket:
Merging now and will followup with @raucao as it's not related to this PR.

PR #324, opened 2021-02-12
ignore package-lock.json

silverbucket:


raucao:
Wondering what the reason is...

silverbucket:
> Wondering what the reason is...

Everything uses `yarn` now, but sometimes I forget and run `npm`, this just keeps the `git status` less noisy.

raucao (approved):
:+1:

raucao:
Actually, the `package-lock.json` I get from `npm install` only contains the dev dependencies like lerna and yarn. I think it's probably not a bad idea to fix the versions for those. Other than that, the `npm install` runs `lerna bootstrap` anyway.

silverbucket:
@raucao it's not for the root package.json (which, as you say, uses lerna commands) but for each of the packages/*/package.json. both yarn and npm use the package.json file the same way, so forgetting to run yarn and running npm instead will create a package-lock.json

raucao:
> so forgetting to run yarn and running npm instead will create a package-lock.json

You run `npm install` to install the Yarn dependency in the root folder. That's `the package-lock.json` I'm talking about.

silverbucket:
> You run `npm install` to install the Yarn and Lerna dependencies in the root folder. That's `the package-lock.json` I'm talking about.

Yes, this addition to the .gitignore is for `**/package-lock.json`

raucao:
Be that as it may, the root folders lockfile is not tracked in the Git repo for me.

silverbucket:
> Be that as it may, the root folder's lockfile is not tracked in the Git repo for me.

The root folders lock file should be `yarn.lock` which is tracked. If `package-lock.json` shows up there (eg. running `npm install <modulename>`) it should be ignored by this `.gitignore` addition.

raucao:
How can installing yarn using npm create a yarn lockfile? I'm talking about the intentional `npm install` lockfile that locks the yarn and lerna versions from `npm install`.

silverbucket:
> How can installing yarn using npm create a yarn lockfile? I'm talking about the intentional `npm install` lockfile that locks the yarn and lerna versions from `npm install`.

Ah, I see what you're getting at. All dependencies, including the yarn version, are installed with yarn. Yarn, the command, is (should be) installed globally using whatever package manager you want. When called it uses whichever version is installed in the node_modules directory if something exists. Otherwise it uses itself to install it. 

Lerna (`lerna.json`) is configured to use yarn. So `npm install` (npm -> package.json -> lerna -> yarn) is synonymous with `yarn install` (yarn -> package.json -> lerna -> yarn) -- both use yarn for package installation and version locking.

raucao:
Ah, now I understand. Was not expecting to need yarn to install yarn. :clown_face:

PR #325, opened 2021-02-12
Consolidate linting and compliance

silverbucket:
Moves the existing sockethub `.eslintrc` file to the root directory, so it applies to all modules instead of just one.


silverbucket:
Weird error with codeclimate `Parsing error: parser.parse is not a function` 
https://github.com/gilbarbara/codeclimate-stylelint/issues/33#issuecomment-778470025

raucao:
It would be useful to keep unrelated changes out of PRs, but especially to keep unrelated changes out of single commits!

raucao (approved):
:+1:

silverbucket:
Ended up being a lot more work debugging travis and making sure all packages tests are run as expected, putting it back up for review as a lot more commits were made since @raucao approved a few days ago.

raucao (approved):
:+1:

PR #327, opened 2021-02-18
Fix socket.io breaking change (CORS issue)

silverbucket:


raucao (approved):
I have confirmed the fix is working.

PR #328, opened 2021-02-18
Remove dist folder from version control

silverbucket:
The dist folder used to have a purpose for being checked in to the repo, but that changed a while back and I just never removed this. So doing it now. The dist folder is now always generated at run time (ts->js) and is also used during package publishing.


silverbucket:
noice!
Screenshot 2021-02-18 at 23 38 30

galfert (approved):
👍

PR #332, opened 2021-02-18
Add debug logging toggle to SockethubClient

silverbucket:
Show replay objects played back to server when SockethubClient instances `debug` property is true


raucao (approved):
:+1:

PR #333, opened 2021-02-18
Ensure config is only loaded once per thread

silverbucket:
This is just to ensure that nowhere the config object is being re-initialized within the same thread, as that could be indicative of a larger issue.


raucao (approved):
:+1:

PR #340, opened 2021-02-20
Add yarn/npm script for ESLint

raucao:
Easier to run for devs and makes the command composable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,9 @@
     "version": "lerna run build && git add **/dist/*",
     "install": "lerna bootstrap",
     "test": "lerna run test",
-    "travis": "eslint \"**/*.ts\" \"**/*.js\" ./packages/sockethub/bin/sockethub && lerna run travis",
+    "lint:js": "eslint '**/*.ts' '**/*.js'",
+    "lint:fix": "eslint '**/*.ts' '**/*.js' --fix",
+    "travis": "yarn run lint:js ./packages/sockethub/bin/sockethub && lerna run travis",
     "build": "lerna run build",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -13,7 +13,9 @@
     "version": "lerna run build && git add **/dist/*",
     "install": "lerna bootstrap",
     "test": "lerna run test",
-    "travis": "eslint \"**/*.ts\" \"**/*.js\" ./packages/sockethub/bin/sockethub && lerna run travis",
+    "lint:js": "eslint '**/*.ts' '**/*.js'",
+    "lint:fix": "eslint '**/*.ts' '**/*.js' --fix",
+    "travis": "yarn run lint:js ./packages/sockethub/bin/sockethub && lerna run travis",
```
silverbucket: @raucao shouldn't the `./packages/sockethub/bin/sockethub` part be in the linting command and not in the travis one?

silverbucket (changes requested):

package.json
raucao: Why would you want to run the Sockethub binary after linting?

silverbucket: @raucao that wasn't actually what it was doing before, it was being linted. The eslint params take files to lint, it catches all `.js` and `.ts` files and the `bin/sockethub` must be specified explicitly since it has no extension.

raucao: Huh, that's weird! Why would there be a file without extension, instead of using the normal npm layout of specifying executables in the `bin` section of `package.json`?

silverbucket: It's also specified in the `bin` section of the `package.json` however the name itself is `sockethub` more for when you've installed it into your system `npm install -g sockethub` ... then you get a `sockethub` command.

silverbucket (approved):

PR #344, opened 2021-03-10
New config option to define which platform methods require the credentials option passed in. 

### LINKED ISSUE #287: Remove passing of credentials on each platform method
Only pass in credentials on `connect`

silverbucket:
Currently credentials are passed in to every platform function which were previously used to access the appropriate client instance or re-connect if needed. The architecture has progressively moved away from that to simply destroying and restarting the platform if the connection is lost. This PR adds a `requireCredentials` property to the platform config, which is an array listing every method that needs the credentials object passed in. For now, all functions are defined, but as the client internals of each platform is improved we can slowly remove methods from the list.
Resolves #287

### DIFF
--- a/packages/sockethub/src/platform.ts
+++ b/packages/sockethub/src/platform.ts
@@ -130,10 +130,18 @@ function startQueueListener(refresh: boolean = false) {
     const sessionSecret = job.data.msg.sessionSecret;
     delete job.data.msg.sessionSecret;
     getCredentials(job.data.msg.actor['@id'], job.data.socket, sessionSecret,
-                   (err, credentials) => {
-                     if (err) { return done(err); }
-                     platform[job.data.msg['@type']](job.data.msg, credentials, done);
-                   });
+      (err, credentials) => {
+        if (err) { return done(err); }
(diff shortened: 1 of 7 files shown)


packages/sockethub/src/platform.ts
```diff
@@ -130,10 +130,18 @@ function startQueueListener(refresh: boolean = false) {
...
-                     if (err) { return done(err); }
-                     platform[job.data.msg['@type']](job.data.msg, credentials, done);
-                   });
+      (err, credentials) => {
+        if (err) { return done(err); }
+        const params = [ job.data.msg ];
+        if ((Array.isArray(platform.config.requireCredentials)) &&
+            (platform.config.requireCredentials.includes(job.data.msg['@type']))) {
+          // pass credentials object as second param if this method is defined in the
+          // `requireCredentials` list in the platform config.
```
galfert:
```suggestion
          // add the credentials object if this method requires it
```

The current comment just repeats exactly what the code is doing :)

galfert (approved):
Looks good to me. Just left a suggestion for changing (simplifying) a code comment.

PR #356, opened 2021-04-13
Pass complete job object to handleJobResult method

### LINKED ISSUE #354: Bull migration throws exceptions
I'm connected with one IRC and two XMPP accounts in Hyperchannel, and with the latest commit I get the logs absolutely littered with this exception:

```plain
(node:112051) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'socket' of undefined                                                                        
    at PlatformInstance.handleJobResult (/home/basti/src/sockethub/sockethub/packages/sockethub/dist/platform-instance.js:133:57)                                            
    at /home/basti/src/sockethub/sockethub/packages/sockethub/dist/platform-instance.js:67:22                                                                                
    at processTicksAndRejections (internal/process/task_queues.js:97:5)  
```

This results in joins not confirming to the client. Unfortunately, it's not clear what exactly is breaking just from reading those messages. On the commit before dea9026, everything still works.

galfert:
Fixes #354

This fixes the error "TypeError: Cannot read property 'socket' of undefined", by passing the complete job object instead of just the decrypted message to the job handler.

### DIFF
--- a/packages/sockethub/src/platform-instance.ts
+++ b/packages/sockethub/src/platform-instance.ts
@@ -85,7 +85,8 @@ export default class PlatformInstance {
     this.queue = new Queue(this.parentId + this.id, redisConfig);
     this.queue.on('global:completed', (jobId, result) => {
       this.queue.getJob(jobId).then((job) => {
-        this.handleJobResult('completed', crypto.decrypt(job.data.msg, secret), result);
+        job.data.msg = crypto.decrypt(job.data.msg, secret);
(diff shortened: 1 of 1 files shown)


silverbucket:
@galfert I also fixed this last night, but then got caught up in improving the typing (which could have prevented this bug), so anyway I'll merge this and submit a different PR for that work. Thanks!

silverbucket (approved):

PR #359, opened 2021-04-13
bugfix: required param to Bulls clean method

silverbucket:


galfert (approved):
Seems to work. What does the number mean?

silverbucket:
> Seems to work. What does the number mean?

The minimum time surpassed for the job, for it to be cleaned. We're wiping them all.

PR #360, opened 2021-04-13
Improve typing

silverbucket:
Needs the update-dependencies PR to be merged first, as this was branched from that one.


galfert (changes requested):
I don't know what's going on, but using Hyperchannel I can't join channels anymore.

The browser log shows unhandled failed events reported by `sockethub-client`:

Screenshot 2021-04-15 at 12 48 31

The event object looks like it should be a normal `join` event instead of a failure. The Sockethub logs don't show any errors.

silverbucket:
@galfert I figured out the first issue, which is that the error was being swallowed up by bull. So even though the failed event was fired for the job completion (which sends the job back on the 'failed' event via socket.io to the client) there was no reason given. 

Bull uses the normal convention of the first param of a callback being an error, however it needs to be an `Error` instance. I knew this but I guess forgot to change that part of the code. So now we're getting the actual error, which is that the credentials are unable to be fetched/decrypted from the store. 

This shouldn't be a difficult fix, but I'm done for the night. I'll try to add a test for it and fix it in the next few days.

raucao (approved):
Code LGTM. utACK.

silverbucket:
ping @galfert -- since you found the original issue, it would be cool if you had a chance to sign off on this now, assuming it all looks good.

galfert (approved):
Tested and everything seems to work fine now. Changes also looks good to me.

PR #362, opened 2021-04-17
Add metadata to incoming XMPP messages

raucao:
Add timestamp for messages that haven't been received live (e.g. from offline storage, or room archives). (This already results in correct timestamps for messages in Hyperchannel.)

https://xmpp.org/extensions/xep-0203.html
https://www.w3.org/TR/activitystreams-vocabulary/#dfn-published

Use unique stanza ID instead of random integers.

https://xmpp.org/extensions/xep-0359.html

The commit also does a bit of drive-by cleanup of the code turning incoming chat messages into AS objects. It also ditches messages without content/body, of which I found some that aren't important and that we can support later if we choose.


silverbucket (approved):
LGTM! Thanks!

PR #364, opened 2021-04-21
Update docker files

silverbucket:


raucao:
The CI actually failed for this one...

silverbucket:
@raucao I don't think so... https://www.travis-ci.com/github/sockethub/sockethub/builds/223746156

raucao:
https://github.com/sockethub/sockethub/runs/2400760782

silverbucket:
@raucao  seems this new dependency management bot is force-pushing in their PRs which causes problems in the CI finding the ref in git. :/

silverbucket:
Oh wait, then why would it be happening here? I think I'm confusing two different issues. My previous comment is valid for the other PRs that have come in today, but not this one. This one is likely due to https://github.com/sockethub/sockethub/pull/373

raucao:
> seems this new dependency management bot is force-pushing in their PRs

That's a terrible thing to do.

silverbucket:
> > seems this new dependency management bot is force-pushing in their PRs
> 
> That's a terrible thing to do.

Yeah a lot of weird things going on with the CI now. Not so happy with this bot may look for another solution

Issue #349, opened 2021-03-13
Merge SharedResources and ResourceManager


silverbucket:
No need for these to be separate files as one manages cleaning up the other, and having two similarly named files just adds confusion

PR #371, opened 2021-04-21
Update to socket.io v4

### LINKED ISSUE #346: Upgrade sockethub to use socket.io 4.0

silverbucket:
Fixes #346 #349 

Makes use of a new socket.io method (fetchSockets) which allows us to query all sockets by ID, removing the need for the `sessionConnections` shared Map, and the shared-resources.ts file altogether. Likewise the services were simplified into one file, and `redisConfig` moved into the config file. (Also, `resource-manager` was renamed to `janitor` which makes its purpose much clearer). 

Overall less files, less code, the +/- doesn't reflect this because more tests were added.


silverbucket:
@galfert @raucao tests passing, ready for review

packages/sockethub/src/sockethub.ts
```diff
@@ -96,16 +95,14 @@ class Sockethub {
     }
 
     log('active platforms: ', [...init.platforms.keys()]);
-    resourceManager.start();
-
-    // start external services
-    this.io = http.start();
+    janitor.start(); // start cleanup cycle
```
galfert: Wouldn't `janitor.clean()` be an even better name?

galfert (changes requested):
The code itself looks ok to me. But when testing with Hyperchannel, I get a 404 for http://localhost:10550/socket.io.js. In the Sockethub logs I see this error:

> Error: ENOENT: no such file or directory, stat '[...]/sockethub/packages/sockethub/node_modules/socket.io-client/dist/socket.io.js'

galfert (approved):
There is still an open question from my first review. Other than that everything looks and works ok now.

silverbucket:
@galfert yeah I will make that change and then merge.

packages/sockethub/src/sockethub.ts
silverbucket: Agreed, it's more thematic! :)

PR #374, opened 2021-04-22
Fix XMPP "observe" AS object

raucao:
Actor and target were switched, and Hyperchannel had a bug that made it work this way.

### DIFF
--- a/packages/sockethub-platform-xmpp/src/incoming-handlers.js
+++ b/packages/sockethub-platform-xmpp/src/incoming-handlers.js
@@ -184,12 +184,12 @@ class IncomingHandlers {
       this.session.sendToClient({
         '@type': 'observe',
         actor: {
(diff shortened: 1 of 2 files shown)


silverbucket:
I'm not sure about this, there is a difference between incoming and out going objects. For example, in the comments for the observe method, and AS Object coming in from the client would have target as the room, (actor is observing the room). 

However, once the room response with an attendance list (which is not the same as the command completing, but a separate object or set of objects received from the connection and sent to the client), the actor would be the room (the rooms response) and the object would be the list of members. 

That's at least how it's been. I believe this change would make the IRC and XMPP platforms behave differently.

raucao:
Looks like I got confused by the `completed` messages and some kind of strange edge case, that caused "space not found by person ID" errors in Hyperchannel. :/

PR #383, opened 2021-05-05
Simplification of routes, with coverage

silverbucket:
@galfert in response to your feedback in the socket.io v4 PR I ended up adding full test coverage to the routes system (including testing for the existence of the files we serve), additionally I consolidated all of the routing logic into a single file and simplified it quite a bit. This should be reviewed and merged before the socket.io PR as it's branched from that one.


galfert (changes requested):
In Hyperchannel I'm getting this error now:

> Access to XMLHttpRequest at 'http://localhost:10550/sockethub/?EIO=4&transport=polling&t=Nb12c27' from origin 'http://localhost:4200' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.

I can see the CORS headers being added to the library files, but it seems that Socket.io itself is missing some configuration for CORS.

In general I think it would be a good idea if on somewhat bigger PRs you do some basic testing in Hyperchannel yourself before requesting reviews. That should greatly reduce the turnaround time.

silverbucket:
@galfert doh! sorry, yeah I was just testing with the local demos, since they are served from sockethub itself cors isn't relevant, will test on hyperchannel.

silverbucket:
@galfert OK this is ready for review again, verified on hyperchannel, and added more tests for routes and the crypto.randToken function

galfert (approved):
Looking good now 👍

PR #422, opened 2021-05-24
Log when a resource is served, and normalize route path

silverbucket:
Tiny PR to log when a resource is served and to use the same resource base path for each resource.


galfert (approved):
LGTM.

PR #424, opened 2021-05-24
Ensure that platform instance queues don't survive once the instance has been destroyed

silverbucket:
It seemed the Bull queues were not being fully destroyed when the platform was. So subsequent re-creations of the platform (eg. same user address) would create a "new" platform instance but the queue listeners from the old instances would be lingering, resulting in multiple firings of global events.


galfert (approved):
Code looks ok to me. Tested with Hyperchannel.

PR #452, opened 2021-06-07
Update renovate config to be less spammy

silverbucket:
This config will:
* Not separate by packages within the monorepo for the same dependency update
* Only create PRs during the weekdays (avoid weekend pileups)
* Wait 3 days after a release before creating the PR
* Auto-update branches when behind master
* Will create an Issue which will act as a dashboard (overview) or all of the existing and pending (not created yet) PRs


galfert (approved):
LGTM

PR #498, opened 2021-07-29
Integration Tests - Github Actions

silverbucket:
This does all the work to spin up a Prosody docker instance, Redis, and Sockethub via Github Actions, then runs the integration tests which currently just connect to the Prosody server via Sockethub.


raucao:
Can we help with making progress here somehow? Seems like the code is getting a bit stale...

silverbucket:
@raucao  - This branch is dependent on the functionality being completed in https://github.com/sockethub/sockethub/pull/596 which has a lot of breaking changes (including a rewritten sockethub client with callback support, and an entirely new approach to the handler issues we faced previously).

The setback was I re-wrote that code twice (there was a first go, which exists as the previous Reference branch) and it turned out to be a huge bit of work, changed a lot of parts, and still needs unit tests. Ideally the work in that branch would be split into a number of smaller PRs but I lost steam with that prospect (as that was the intent after the first reference branch). 

So, I was planning to get some time to add unit tests for that branch and then submit it for review as a `kredits-3` level PR. 

- As for this branch, it seems stale but it's actually not so bad. Dependencies will need to be sorted, but it was really just waiting on the client callback function (so that tests know when they are completed). Then it can be merged essentially as-is, with more elaborate tests in new PRs. This was just to get the basics working.

silverbucket:
@raucao This is finally ready for review!

raucao (approved):
I was able to run this successfully on my machine as is. Nice! :+1:

PR #508, opened 2021-08-03
Unify coverage and set thresholds

silverbucket:
Uses `c8` everywhere (removes some use of `nyc`), and ensures each package that has tests sets thresholds so we don't drop below in the future.


silverbucket:
@galfert @raucao this PR is ready for review, unifies on `c8` for coverage reporting and sets reasonable thresholds for each package.

silverbucket:
ping @galfert or @raucao for review?

galfert (approved):
Works as intended.

PR #516, opened 2021-08-15
Feeds: Fix results not being sent back correctly

silverbucket:
Fixes bug where results where not being propagated back to the client for non persistent platforms (eg. platforms which are shared between any number of users, contain no user-specific info, and can only send data to users as a response to a request). 

Depends on https://github.com/sockethub/sockethub/pull/517 to be merged first.


silverbucket:
@raucao @galfert - this is ready for review as well, can be merged after #517

Issue #491, opened 2021-07-24
XMPP credentials validation: "credentials schema validation failed: /object = Missing required property: server"


raucao:
Server is now optional (which isn't implemented yet), as we discover it via DNS records automatically.

PR #517, opened 2021-08-15
Improved Middleware

silverbucket:
Greatly simplified the middleware logic, and breaks these pieces into stand-alone functions stored in the middleware/ directory. This refactor was done to make use of the Acknowledgements feature of `socket.io` event sending/callback confirmation. 
https://socket.io/docs/v4/emitting-events/#Acknowledgements

However after the refactor, it seems the feature in socket.io was not working as described in the docs. 
Ticket created here: https://github.com/socketio/socket.io/issues/4057 

- [x] Refactor and split up all middleware to stand alone functions in the `middleware/` directory
- [x] Update jsdoc
- [x] Add unit tests for middleware.chain
- [x] Export middlewares from middleware file
- [x] Add unit tests for createActivityObject middleware
- [x] Add unit tests for storeCredentials middleware
- [x] Add unit tests for validate middleware
- [x] Add unit test to verify exception is not thrown by middleware chain 
- [x] Fix bug where schema middleware chain throws error from middleware function, rather than sending an error object to client.

Created a ticket for confirmation callbacks https://github.com/sockethub/sockethub/issues/519 which will be added once the `socket.io` issue is resolved.


silverbucket:
ping @raucao  or @galfert  when you have time to review, this PR has been ready for review for a while.

raucao:
Ah, one more thing from testing with Hyperchannel: it now actually crashes Sockethub with an uncaught exception when the credential validation fails:

```plain
  sockethub:validate:iw8R5mmzEeXtxcexAAAB applying schema validation for credentials +0ms

UNCAUGHT EXCEPTION

Error: credentials schema validation failed: /object = Missing required property: server
    at processCredentials (/home/basti/src/sockethub/sockethub/packages/sockethub/dist/middleware/validate.js:100:21)
```

(The validation there is outdated, with #491 open to update it.)

silverbucket:
> Or even better:
> 
> ```js
> import * as middleware from 'middleware';
> 
> middleware.validate();
> middleware.storeCredentials();
> ```

Good point, I agree this looks best, I was trying to keep the files small and self-contained, but the self-documenting nature of this example wins IMO. 



> Ah, one more thing from testing with Hyperchannel: it now actually crashes Sockethub with an uncaught exception when the credential validation fails:

Nice catch! The previous validation code was careful about bubbling up errors without leaking exceptions from "known" error (like schema validation). Something must have slipped through in the refactor so I'll try to add tests to catch this then fix it.

packages/sockethub/src/middleware/validate.test.data.js
```diff
@@ -0,0 +1,286 @@
+export default [
+  {
+    "name": "basic",
+    "result":true,
```
raucao: How about this?

```suggestion
    "valid":true,
```

packages/sockethub/src/middleware/validate.test.ts
```diff
@@ -0,0 +1,32 @@
+import { expect } from 'chai';
+
+import validate from './validate';
+import asObjects from "./validate.test.data.js";
+import ActivityStreams from 'activity-streams';
+
+const activity = ActivityStreams();
+
+describe('Validates given AS object', () => {
```
raucao:
```suggestion
describe('AS object validations', () => {
```

packages/sockethub/src/middleware/validate.test.data.js
silverbucket: @raucao yeah, I like that better too

silverbucket:
ping @raucao @galfert ^^ :)

raucao (approved):
Code LGTM, and works with Hyperchannel. :+1:

PR #518, opened 2021-08-16
Fix polynominal time required to match regexp

silverbucket:
Fix for what LGTM reported as a polynominal regexp, more info here: https://lgtm.com/rules/1511427506129/


galfert (approved):
👍

Issue #486, opened 2021-07-18
Enable source maps for sockethub-client.js


raucao:
Currently seeing the following error in my browser console with `4.0.1`. Would be nice if source maps worked out of the box...

```plain
Source map error: Error: request failed with status 404
Resource URL: http://localhost:10550/sockethub-client.js
Source Map URL: client.js.map
```

PR #521, opened 2021-08-16
Add route to generated source map for sockethub-client

silverbucket:
Resolves https://github.com/sockethub/sockethub/issues/486

@raucao can you verify this works for you?


raucao (changes requested):
It looks llike there's a URL mismatch still (check the filename):

```
Source map error: Error: request failed with status 404
Resource URL: http://localhost:10550/sockethub-client.js
Source Map URL: client.js.map
```

silverbucket:
@raucao OK, it's a bit of a hack but it should work now. I was able to strip away full path references, but have been unable to remove relative paths. So it specifies the path as `js/sockethub-client.js.map` even though the original JS file is served from `/sockethub-client.js`. I've added a second route to server the map from either `/` or `/js/`, and this works. Not pretty, but it works. :)

raucao:
I'm not sure what you're testing this with, but I still get the error:

```plain
Source map error: Error: request failed with status 404
Resource URL: http://localhost:10550/sockethub-client.js
Source Map URL: client.js.map
```

Do you not see the same? If you use Chrome, then it looks like this for example:

```plain
DevTools failed to load source map: Could not load content for http://localhost:10550/client.js.map: HTTP error: status code 404, net::ERR_HTTP_RESPONSE_CODE_FAILURE
```

silverbucket:
> I'm not sure what you're testing this with, but I still get the error:
> 
> ```
> Source map error: Error: request failed with status 404
> Resource URL: http://localhost:10550/sockethub-client.js
> Source Map URL: client.js.map
> ```
> 
> Do you not see the same? If you use Chrome, then it looks like this for example:
> 
> ```
> DevTools failed to load source map: Could not load content for http://localhost:10550/client.js.map: HTTP error: status code 404, net::ERR_HTTP_RESPONSE_CODE_FAILURE
> ```

Weird, no I don't get this - I get the "Source Map Detected" alter in the devtools when inspecting the code for sockethub-client.js

silverbucket:
@raucao make sure you rm -rf packages/sockethub/dist first - there could be leftover files in there.

silverbucket:
@raucao any update on this? Making sure the old build files are gone should be the fix, because your error is referencing things that don't exist anymore. (client.js is no longer a file)

Issue #495, opened 2021-07-25
sockethub-client overwriting `socket.on` with a version that doesn't support multiple listeners


raucao:
Exactly as stated in the title, it's not possible to register multiple listeners to the same event name with the current `sockethub-client`. It seems to want to do some additional ActivityStreams parsing, but in my use case there was actually no difference between the emitted parsed events and what I got when circumventing the issue by using the `socket.onAny` API for the additional listener instead.

silverbucket:
You are correct that there is some middleware which parses the ActivityStreams, its main purpose is to ensure certain properties exist, expand and collapse objects. For example:

You can register an object once:
```js
sc.ActivityStreams.Object.create({
  '@id': 'foo@kosmos.org`,
  '@type': "person",
  displayName: "Mr.Foo"
});
```

And any time it's sent or received with the ID string:
```js
{
 actor: 'foo@kosmos.org',
 ...
}
```

It would automatically be expanded:
```js
{
  actor: {
    ''@id': 'foo@kosmos.org`,
    '@type': "person",
    displayName: "Mr.Foo"
  }
}
```
Additionally, there is replay functionality in the sockethub client, so it can replay unconfirmed commands when needed (this greatly improved the reliability of keeping state of rooms joined, messages sent, etc.). 

But yeah, I didn't make the new handlers flexible enough, looks like there's just one handler for each unit event name: https://github.com/sockethub/sockethub/blob/8200286125bea0638bbc87c4168cadc1ac4c2eeb/packages/sockethub/src/js/client.js#L148-L158

Should be a fairly simple fix, will try to get to this soon.

PR #522, opened 2021-08-16
Keep client event handlers stored in an array

silverbucket:
Allows for multiple handlers to be registered for the special `on` events that `sockethub-client` needs to intercept. Resolves https://github.com/sockethub/sockethub/issues/495


packages/sockethub/src/js/client.js
```diff
@@ -86,7 +70,9 @@
         } else if (event === 'disconnect') {
           this.online = false;
         }
-        this.__handlers[event](obj);
+        for (let cb of this.__handlers[event]) {
```
raucao, line 73: Looks to me like this should be `const` instead of `let`?

```suggestion
        for (const cb of this.__handlers[event]) {
```

packages/sockethub/src/js/client.js
```diff
@@ -169,7 +155,9 @@
   SockethubClient.prototype.__unpackAndCallHandler = function (event) {
     return (obj) => {
       let unpackedObject = this.ActivityStreams.Stream(obj);
-      this.__handlers[event](unpackedObject);
+      for (let cb of this.__handlers[event]) {
```
raucao, line 158:
```suggestion
      for (const cb of this.__handlers[event]) {
```

raucao (commented):
Yay! :clap: 

Just a syntax nitpick for now (which your text editor should actually warn about, if it's linting the source code live). Have yet to test it with running code.

silverbucket:
I've done some manual testing, but not extensively. I don't have any unit tests set up for this JS file, but plan to add integration tests which would register multiple handlers, etc. (Once that branch work is complete). Created issue: https://github.com/sockethub/sockethub/issues/525

raucao:
Just one question, as I was looking at it: did you consider `removeListener()` to work with this? Because that's actually what we need in Hyperchannel, too.

silverbucket:
@raucao that's not implemented, I'm thinking maybe this would be better to include an actual event emitter which proxies for the socket even emitter, rather than a makeshift one. It would need to basically be re-written though. Maybe this could be a temporary fix for the specific issue with a new task to re-write the whole thing?

raucao (approved):
OK, let's do it as you propose. I was able to test this one with Hyperchannel successfully (albeit the channel join failures directly after the connect still persist, so that's a different issue in the XMPP platform itself).

silverbucket:
Created https://github.com/sockethub/sockethub/issues/526

PR #523, opened 2021-08-16
Fix broken linting

silverbucket:
Linting checks weren't being run properly during compliance, fixed it and fixed some linting errors.


raucao (approved):
LGTM

PR #524, opened 2021-08-16
Remove support for node 15

silverbucket:
Removed support for node 15 from compliance checks. ~~Added node 12 as it's still officially supported.~~


raucao (approved):
:+1:

PR #533, opened 2021-08-21
Documentation link fixes

Cobertos:
Fixes a couple documentation relative links


raucao:
I wonder if these paths are intentional so that it works on https://www.npmjs.com/package/sockethub. I wonder how other monorepos deal with this.

Cobertos:
Hmm, I've had this issue before with non monorepos and just had to use an
absolute URL

silverbucket (approved):

PR #538, opened 2021-08-25
Fix redis options not passing to bull#Queue

### LINKED ISSUE #191: config.REDIS is ignored.
No matter how you configure sockethub, 127.0.0.1 is always used as redis server. PR incoming, by replacing each redisPool with a shared redisPool attached to the sockethub class. May not be ideal; I would love some input.

EDIT: for knowledge, I'm using sockethub as part of working through https://unhosted.org/adventures/6/Controlling-your-server-over-a-WebSocket.html.

Cobertos:
The default export of config does not have any keys except `get()`. Therefore `bull#Queue` was not getting any redis config and using defaults instead

Fixes #191

This is currently only broken in `v4.0.1`. Wasn't able to target the tag


packages/sockethub/src/platform-instance.ts
```diff
@@ -74,7 +74,7 @@ export default class PlatformInstance {
    * When jobs are completed or failed, we prepare the results and send them to the client socket
    */
   public initQueue(secret: string) {
-    this.queue = new Queue(this.parentId + this.id, redisConfig);
+    this.queue = new Queue(this.parentId + this.id, { redis: config.get('redis') });
```
silverbucket, line 90: We need a way to ensure proper use of the `createClient` call, we should verify if we are setting things up wrong or otherwise determine if our implementation is correct and this is a bull issue (perhaps?)

silverbucket:
Possible related Bull issue https://github.com/OptimalBits/bull/issues/841

silverbucket:
After thinking about this a while, it's clear that `createClient` for some reason isn't working as expected, and it will take more investigate to get right -- as there will need to be some manual management of Redis connections. Therefore it probably make sense to simplify the use of Redis to what we know works (this PR), and create a ticket for further investigate to improve possible scalability of Redis connections and Bull. We'll need both an approach for implementation but also for properly verifying behavior.

Created https://github.com/sockethub/sockethub/issues/563

silverbucket (approved):

PR #553, opened 2021-08-29
XMPP credentials `server` property no longer required

### LINKED ISSUE #491: XMPP credentials validation: "credentials schema validation failed: /object = Missing required property: server"

silverbucket:
Fixes #491


raucao (approved):
LGTM

Issue #355, opened 2021-04-08
Add eslint plugin for node security


silverbucket:
https://www.npmjs.com/package/eslint-plugin-security-node

PR #554, opened 2021-08-29
Added linting checks for common node security vectors

silverbucket:
Resolves https://github.com/sockethub/sockethub/issues/355


raucao (approved):
:+1:

PR #558, opened 2021-08-30
Ensure actor on error objects when possible

### LINKED ISSUE #492: Connect error AS object should contain actor/target
E.g. a "not-authorized" comes back without knowing which account failed to authorize.

silverbucket:
Resolves #492


raucao:
Great!

Can we also ensure that `failure` messages do not create additional error `message`s that contain the same data?

packages/sockethub/src/platform-instance.ts
```diff
@@ -180,7 +184,7 @@ export default class PlatformInstance {
     const errorObject: ActivityObject = {
       context: this.name,
       '@type': 'error',
-      target: this.actor,
+      actor: this.actor,
```
raucao, line 188: I'm wondering if `target` doesn't make more sense here, as the entity sending the error is not the user, but usually a server?

silverbucket: Yarg! I don't know :D because in the case of a failed job, it's sending back the job object (actor is user) with the error message... additionally, in this case, it's sending the error object back as a message... so either that can both have the actor=user, or one can be actor=user and the other can be target=user ... does that make more sense? Or I could just remove this message altogether, though then hyperchannel wouldn't receive the failure since they aren't getting the failure message.

raucao: Hyperchannel would then actually receive the failure message, because it's not being superceded by the message message which triggers the `offAny`.

> I don't know :D because in the case of a failed job, it's sending back the job object (actor is user) with the error message

Yeah, I'm also not sure, hence the question. The IRC connect failure sends a target with the user info, so at least it should be consistent between platforms then.

silverbucket: The more I've been thinking of this the more a few things keep coming back to the forefront in Sockethubs expectations, eg. an actor property should always be  present. In this case, I think it makes sense that the actor is still the user@host because the report is that `user@host` generates the error. 

Additionally, to keep context and naming sensible, the PlatformInstance is what's assigning the `actor` based on its `actor` property, and to swap the actor to the target seems more confusing going forward. 

Finally, there's the idea of unsolicited/non response messages being the ones where the target is the user, and the actor is some other entity. I don't think necessarily it makes things more clear to swap these for error messages. 

So I'm thinking we keep things simple, stick with ensuring an actor if not present, and leave it at that (fixing the IRC one). What do you think?

raucao: OK, let's do it like this then!

raucao:
I believe this question is still open?

> Could we also ensure that `failure` messages do not create additional error `message`s that contain the same data perhaps? Or is this intentional, so that one is for the software to read, and the other one actually for the user?

raucao (changes requested):
Hmm, with this branch I'm now getting all error messages twice from Sockethub:

Screenshot from 2021-09-01 18-41-27

raucao (approved):
As @galfert verified on chat, the double-message issue is a new bug in Hyperchannel. So I think this is ready to merge.

silverbucket:
@galfert @raucao - OK, I've simplified things:

1. Any job failures, callback is used. In XMPP I was using both the cb() and sendToClient, for no reason other than to be confusing I guess.
2. Platform.ts ensures that the callback can only be used once (ignores subsequent calls)
3. Don't throw an error when not necessary, instead use fail callback.

Can I get a final OK just to be sure?

raucao:
OK, so now I found out why Hyperchannel isn't logging the failure messages for the connect. Instead of the normal "failure" event name, which other failures do use, this one uses the event name "failed".

I think we need to manually ensure that these are all consistent for now. But in the future, trying to send unknown event names should probably cause at least a console warning, or ideally a linting error.

PR #559, opened 2021-08-30
Update renovate config

silverbucket:
* Mondays only
* 4 days of package stability before PR is created
* Removed old jest version pegging
* Update label name to `type:dependencies`


galfert (approved):
I thought we talked about doing something like an "Update tuesday", but if that's how you want to start your week, go ahead :)

PR #560, opened 2021-08-30
Github Actions: split linting and unit tests

silverbucket:
In an attempt to get rid of the ghost entry for node v15 in the Github Actions checks, (and also to improve visual reporting in the list), I've split the linting and unit tests into separate checks to run.


silverbucket:
Ugh, I have no idea why this `build (15.x)` is still here, these are completely new github actions workflow files, no mention of 15.x anywhere, just it still shows up in the list of checks. I've tried asking here with no luck so far: https://stackoverflow.com/questions/68973531/removed-node-js-v15-from-list-of-versions-but-github-actions-still-shows-it-in

Anyone have any ideas? It seems to be something in the Github (website) settings rather than in the repository config.

silverbucket:
Ah! Finally got it, these ghost entries were showing up due to previously being set as required. So, in the repository settings -> branches -> master [edit] page, you can remove those by name and add the new checks by name. It's too bad this isn't more obvious. Anyway, it's finally gone.

raucao (approved):
thumbs-up-asian-guy-gif

PR #561, opened 2021-08-30
Fix build and dist files

silverbucket:
Leftover `dist/` files (eg. no longer built but still exist in the directory) sometimes cause an issue to not be caught right away. This fixes an issue where a file was no longer being generated to `dist/` but it wasn't caught in the PR that introduced the change.

Now we clean the `dist/` dir every time, and the `routes.test.ts` tests for the existence of dist files, rather than the src files.


galfert (approved):
Looks ok to me.

raucao:
The code here looks more like a `kredits-1` to me tbh...

silverbucket:
> The code here looks more like a `kredits-1` to me tbh...

It took a while to figure out the cause and figure out and apply a useful fix, even if few lines were changed. As I understand it, `kredits-2` is for most normal work chunks, unless it is plainly simple/obvious or exceedingly time consuming.

Issue #564, opened 2021-08-31
"unavailable" presence stanzas interpreted as "online" updates


galfert:
An incoming stanza for a user leaving a room is sent to the Sockethub client as an "online" presence update.

Example for the incoming stanza:

```json
{
    "attrs": {
        "from": "kosmos-random@kosmos.chat/galfert",
        "to": "jimmy@kosmos.org/hyperchannel",
        "type": "unavailable",
        "xml:lang": "en"
    },
    "children": [
        {
            "attrs": {
                "xmlns": "http://jabber.org/protocol/muc#user"
            },
            "children": [
                {
                    "attrs": {
                        "affiliation": "owner",
                        "role": "none"
                    },
                    "children": [],
                    "name": "item"
                }
            ],
            "name": "x"
        }
    ],
    "name": "presence"
}
```

The object that Sockethub sends to the client:

```json
{
    "@type": "update",
    "actor": {
        "@type": "person",
        "@id": "kosmos-random@kosmos.chat/galfert"
    },
    "object": {
        "@type": "presence",
        "status": "",
        "presence": "online"
    },
    "target": {
        "@id": "jimmy@kosmos.org/hyperchannel"
    },
    "context": "xmpp"
}
```

silverbucket:
@galfert @raucao - does this look like a reasonable looking XML representation of what comes in?

```xml
<presence to="foo@bar.org/hyperchannel" from="baz@bag.org" type="unavailable">
  <x xmlns="http://jabber.org/protocol/muc#user">
    <item affiliation="owner" role="none"></item>
  </x>
</presence>
```

Here's what it *should* output:

```js
{
  "@type": "update", 
  actor: { "@type": "person", "@id": "baz@bag.org" },
  target: { "@id": "foo@bar.org/hyperchannel" },
  object: { "@type": "presence", "status": "", "presence": "unavailable" },
}
```
This is what I'll use for the unit test data in incoming-handlers, let me know if I got anything wrong.

silverbucket:
@galfert OK, PR updated and ready for review

PR #565, opened 2021-08-31
XMPP Presence `unavailable`

silverbucket:
Adds support for the XMPP presence type `unavailable`.
Resolves https://github.com/sockethub/sockethub/issues/564


galfert (approved):
Confirmed it to work now 👍

PR #572, opened 2021-09-03
Split validate logic into two functions: expandActivityStream & validate

silverbucket:
Previously the validation process both expanded the incoming AS object and validated it against schemas. It was convoluted and messy. 

This PR clearly separates the expanding of the object (eg.  props like `actor:"foo"` to `actor:{"@id":"foo"}`) and the validation (schemas, context value, etc. into two middleware functions.


silverbucket:
@raucao @galfert - this PR is ready for review, along with 2 others (which are branched off each other), so this one (#572) followed by #573 then #574

PR #573, opened 2021-09-03
Switch to from `tv4` to `ajv`, update schemas. 

silverbucket:
The JSON schema validation tool `tv4` hasn't been updated in years, and still only supported draft 04. It seems, these days, `ajv` is the popular and active schema validation tool and supports all the newest drafts. 

I was having weird issues validating some aspects of the schemas with `tv4` which prompted the switch (which needed to happen eventually anyway). 

Depends on #572 to be merged first.


packages/sockethub-schemas/tests/schemas.test.data.js
```diff
@@ -0,0 +1,173 @@
...
+      object: {
+        '@type': 'credentials'
+      }
+    },
+    false,
+    '/target: must have required property \'@id\''
+  ],
+
+  [
+    'bad iri in target',
```
galfert: I think the description is wrong. The object doesn't even have a `target`.

packages/sockethub-schemas/tests/schemas.test.data.js
```diff
@@ -0,0 +1,173 @@
...
+        '@id': 'example.org/some/path.html',
+        '@type': 'website'
+      }
+    },
+    false,
+    '/object/@type: must be equal to one of the allowed values'
+  ],
+
+  [
+    'valid iri in target',
```
galfert: Same as above. There is no `target`.

packages/sockethub-schemas/tests/schemas.test.data.js
```diff
@@ -0,0 +1,173 @@
...
+        displayName: 'bob'
+      },
+      object: {
+        '@type': 'feed',
+        '@id': 'http://rss.example.org/feed.rss'
+      }
+    },
+    false,
+    '/actor: must have required property \'@id\''
+  ], [
```
galfert: Missing an empty line here to improve readability.

packages/sockethub/src/bootstrap/platforms.js
```diff
@@ -66,28 +67,23 @@ module.exports = function platformLoad(moduleList) {
         const packageJson = require(path + '/package.json');
         let types = [];
 
+        const validate_platform_schema = ajv.compile(schemas.platform);
         // validate schema property
-        if (! tv4.validate(p.schema, schemas.platform)) {
+        if (! validate_platform_schema(p.schema)) {
           throw new Error(
-            `${platformName} platform schema failed to validate: ${tv4.error.message}`);
+            `${platformName} platform schema failed to validate: `, ajv.errors);
```
galfert: Can you just add the `ajv.errors` as second argument when creating a new `Error` object?

If this is the standard JS `Error`, then the second argument needs to be an object with a `cause` property: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error

packages/sockethub/yarn.lock
```diff
@@ -904,6 +904,7 @@ acorn@^8.4.1:
   resolved "https://registry.yarnpkg.com/acorn/-/acorn-8.4.1.tgz#56c36251fc7cabc7096adc18f05afe814321a28c"
   integrity sha512-asabaBSkEKosYKMITunzX177CXxQ4Q8BSSzMTKD+FefUhipQC70gfW5SiUDhYQ3vk8G+81HqQk7Fv9OXwwn9KA==
 
+<<<<<<< HEAD
```
raucao: Oops...

raucao (changes requested):
Latest additions are merge conflicts.

packages/sockethub/yarn.lock
silverbucket: damn lockfiles! fixed.

packages/sockethub/src/middleware/validate.test.data.ts
```diff
@@ -260,5 +258,51 @@ export default [
...
+  },
+  {
+    "name":"traditional message",
+    "valid":true,
+    "type":"message",
+    "input":{
+      "@type": "foo",
+      "context": "irc",
+      "actor": { "@id": "irc://uuu@localhost", "@type": "person" }
+    }
```
galfert, line 306: If I see correctly, the tests for failing validations are also testing the actual error message. Then there might be a test missing for this error https://github.com/sockethub/sockethub/pull/573/files#diff-2e9c7a22f756a8dfb626c2786ace21077766c34452d8f439a2638916a8eeff2eR74

galfert (commented):
Left a few comments/questions.

packages/sockethub/src/middleware/validate.test.data.ts
silverbucket: Well, while it uses the functionality of platform.js it's technically not testing that platformLoad method. I am avoiding writing tests for the JS files in bootstrap. ;)

galfert (approved):
Looks all ok to me now.

PR #574, opened 2021-09-03
Update AS2.0 names

### LINKED ISSUE #215: AS2.0 property naming updates
* `@id` is now `id`
* `objectType`, `verb` or `@type` is now `type`. 
* `displayName` is now `name`
* `post` split to `create` & `add`
https://www.w3.org/TR/activitystreams-core/

silverbucket:
This PR finally switches everything over to AS2.0 names for some of the required properties sockethub uses for routing data (id and type specifically).

* `@id` -> `id`
* `@type` -> `type`
* `displayName` -> `name`

Resolves #215 

Depends on #573 to be merged first.


silverbucket:
BTW - It should be noted that the `activity-streams.js` module was also updated to include the former names in its renaming table. So existing objects which have `@type` and `@id` etc. will be mapped to the new names `type` `id`.

The only questionable piece is that on the root of AS2.0 objects you can still have `@type` and I believe also `@id` which mean different things than those properties without the `@`. `activity-streams.js` currently does not know how to make any sort of distinction.

raucao (approved):
I have checked the vocabulary spec and this LGTM in regards to that.

galfert:
> BTW - It should be noted that the `activity-streams.js` module was also updated to include the former names in its renaming table. So existing objects which have `@type` and `@id` etc. will be mapped to the new names `type` `id`.

That doesn't seem to work. When using SH master with Hyperchannel, I see a bunch of schema validation errors (e.g. "failure handling credentials. Error: activity stream must contain a type property.") in the SH logs.
And in Hyperchannel I just get failure events with an empty error object.

silverbucket:
@galfert Can you show what the credential object you are sending looks like? (Maybe we can take this to a new issue?)

PR #606, opened 2021-10-05
Renovate Config

silverbucket:
Main change here is that we will now group patch and minor releases in the devDependencies group together into one PR.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,8 +1,15 @@
 {
   "labels": [ "kredits-1", "type:dependencies" ],
-  "reviewers": [ "silverbucket" ],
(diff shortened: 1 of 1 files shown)


silverbucket:
@raucao @galfert - needs review, this should reduce dependency update spam even more by grouping all devDependency updates that are patch or minor into one PR

.github/renovate.json
```diff
@@ -1,14 +1,22 @@
 {
   "labels": [ "kredits-1", "type:dependencies" ],
-  "reviewers": [ "silverbucket" ],
+  "reviewers": [ "silverbucket", "galfert", "raucao" ],
```
raucao:
```suggestion
  "reviewers": [ "silverbucket", "galfert" ],
```

PR #644, opened 2021-12-13
Remove old typing for kue

silverbucket:
Somehow this typing was overlooked and stuck around after we switched from kue to bull.


raucao:
Do you think the kredits here make sense, considering that the PR that removed kue and should have contained this line change was properly kredited already? I usually just push changes like that directly to master and/or merge immediately myself without adding a label.

raucao:
I think there is "total settling" a social mechanism like this upfront. Any and all rules and guidelines are always open to interpretation, no matter how clearly defined. We definitely need to write down the guidelines however, because I'm not aware we even have documentation for that in the first place, since we only experimented with it for a while now and are still learning what works and how.

Formatting wiki entries does not automatically issue kredits, because the wiki has the "minor changes" option for that, which I regularly use for anything that's not an actual change in content. It also doesn't issue 500 kredits for every change, but only proposes 500 for all changes combined within 24 hours that are not minor.

silverbucket:
@raucao Ah I see, I think I forgot/misremembered about that setup for the Wiki. Makes a lot more sense, and will keep that in mind. I agree that a quick reference guide would be super helpful.

PR #645, opened 2021-12-13
Update renovate.json

silverbucket:
Should update once a month, putting all patch and minor releases into one PR, but still separating major releases as these should really be tested separately to properly identify breaking changes, if any.


raucao (approved):
LGTM :+1:

galfert:
Cool. Is there also a config to create a merge commit as well? Or did you just use the "Rebase and merge" button before?

silverbucket:
@galfert I didn't see anything that wasn't related to auto-merging, but also- to be honest I don't fully understand the difference between a rebase and a merge commit', so I just searched for 'merge commit' and that was all I found. Maybe you would find something appropriate though? https://docs.renovatebot.com/configuration-options/

raucao:
It's "rebaseWhen":

> Renovate will autodetect the best setting. Defaults to conflicted unless the repository has a setting requiring PRs to be up to date with the base branch

That setting is set for the master branch in this repo. We can either remove the branch setting, or set "rebaseWhen" to "never" AFAICS.

galfert:
@silverbucket I think it's not a renovate config, but rather you pressing the "Rebase and merge" button when merging PRs, instead of "Create a merge commit":

Screenshot 2021-12-13 at 18 19 58

When using the rebase function, all commits of the PR are just put on top of master, so the history is just one straight line of commits:

Screenshot 2021-12-09 at 19 11 19

When using a merge commit, the history gains a bit more context. The commits of a PR are grouped together and the merge commit itself references the PR number in the commit message and the person that merged it. That way you can always look it up on Github again to see the context and discussions.

Also, and I think that's the important part in regards to the renovate issue, you can more easily skip a whole bunch of commits when looking at the history, because you can see they all belong to one PR.

Screenshot 2021-12-09 at 19 12 00

galfert:
> It's [rebaseWhen](https://docs.renovatebot.com/configuration-options/#rebasewhen):
> 
> > Renovate will autodetect the best setting. Defaults to conflicted unless the repository has a setting requiring PRs to be up to date with the base branch
> 
> That setting is set for the master branch in this repo. We can either remove the branch setting, or set "rebaseWhen" to "never" AFAICS.

No, I think that just configures if renovate rebases an open PR automatically after other PRs have been merged.

I noticed that my PR that was merged earlier also doesn't have a merge commit.

silverbucket:
@galfert Ah I see, OK I will try to remember to do that with renovate PRs from now on. It would be great if there was a way to set that to the default just for those though. Because, as I understand, we don't normally want to do that for use PRs right?

galfert:
No, I would prefer it for all PR merges.

silverbucket:
Ok, I'll update that setting

PR #648, opened 2021-12-13
Update irc2as to use new property names

silverbucket:


silverbucket:
@galfert this should be ready to review

galfert (approved):
Awesome. Everything's working now 👍

PR #649, opened 2021-12-24
Support incoming message correction for XMPP messages

raucao:
* Uses the message ID as default, and adds the stanza ID in a separate, new property
* Adds another new property for message correction/replacement as per XEP-0308

This introduces a breaking change (albeit only for clients who already cache/store messages by ID), in that the main message "id" for XMPP messages has been changed from using the stanza ID to using the message's "id" attribute value. The stanza ID is now a separate property, as it is also needed (for example for MAM).

### DIFF
--- a/packages/activity-streams.js/yarn.lock
+++ b/packages/activity-streams.js/yarn.lock
@@ -8,9 +8,9 @@
   integrity sha512-0hYQ8SB4Db5zvZB4axdMHGwEaQjkZzFjQiN9LVYvIFB2nSUHW9tYpxWriPrWDASIxiaXax83REcLxuSdnGPZtw==
 
 "@discoveryjs/json-ext@^0.5.0":
-  version "0.5.5"
-  resolved "https://registry.yarnpkg.com/@discoveryjs/json-ext/-/json-ext-0.5.5.tgz#9283c9ce5b289a3c4f61c12757469e59377f81f3"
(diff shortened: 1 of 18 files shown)


galfert:
Can confirm that it works with Hyperchannel. But I see warnings in the console:

> activity-streams.min.js:2 invalid property: "xmpp:stanza-id"

Looks like the necessary changes to the schema are missing here.

silverbucket:
> But I see warnings in the console:
> 
> > activity-streams.min.js:2 invalid property: "xmpp:stanza-id"
> 
> Looks like the necessary changes to the schema are missing here.

Yeah this property should be added to `activity-streams.js`. That library may be totally removed eventually but for now best to keep it up to date. Otherwise this looks good to me.

raucao:
I just had a look at adding the properties to the validation. However, I don't want to add specific properties, but only the namespace, so that any properties in the namespace are valid. However, in order to do that correctly, we also need to set `@context`, and so this is quickly adding new features to `activity-streams.js` which shouldn't really be in there, because they are not specific to AS.

> That library may be totally removed eventually

If this is already the rough plan, then maybe now would be a good time? We're going to break the AS format anyway in the next release.

silverbucket:
> > That library may be totally removed eventually
> 
> If this is already the rough plan, then maybe now would be a good time? We're going to break the AS format anyway in the next release.

Something still needs to be done to replace a portion of the functionality it provides (expanding AS objects that are referenced just by ID), so it's not ready to be completely removed yet, but I do plan to remove it as part of the breaking release. So then lets just leave it as-is with the warning, for now.

silverbucket (approved):

PR #653, opened 2022-01-03
Support outgoing message correction

raucao:
As per XEP-0308, and with the same AS object format as incoming corrections are forward to clients.

(Note: the XML fakes are still not tested correctly. I added FIXME comments, but since I wasn't able to drive-by-fix them easily, and they're already broken in master since before I worked on message corrections, I consider the problem out of scope for these feature PRs.)

### DIFF
--- a/packages/sockethub-platform-xmpp/src/index.test.js
+++ b/packages/sockethub-platform-xmpp/src/index.test.js
@@ -76,8 +76,8 @@ const job = {
       object: {
         'type': 'message',
         id: 'hc-1234abcd',
-        replace: { id: 'hc-234bcde' },
-        content: 'hi yall'
+        content: 'hi yall',
+        'xmpp:replace': { id: 'hc-234bcde' }
       },
       target: target.roomuser
     }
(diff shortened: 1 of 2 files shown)


silverbucket:
@raucao Could you maybe create an issue and explain what you mean by wrong message? Do you mean any arbitrary data passes or that something specific about the type?

raucao:
> Do you mean any arbitrary data passes

Yes.

packages/sockethub-platform-xmpp/src/index.test.js
```diff
@@ -67,7 +67,17 @@ const job = {
...
+        content: 'hi all'
+      },
+      target: target.roomuser
+    },
+    correction: {
+      actor: actor,
+      object: {
+        'type': 'message',
+        id: 'hc-1234abcd',
+        replace: { id: 'hc-234bcde' },
```
raucao: Actually, I think this is `xmpp:replace` in the other direction.

image

raucao: Fixed.

raucao:
I have fixed the property name. Should be good to merge now.

raucao:
Can we merge this now?

silverbucket (approved):

silverbucket:
lgtm!

PR #671, opened 2022-01-31
Decouple connect and join steps in IRC platform

### LINKED ISSUE #529: Decouple connect and join steps in IRC platform
I'm not sure if this is new, or how it was before, because Hyperchannel just sends channel joins without even caring about the server being connected first. But now that I'm implementing it properly, I'm seeing that there's just no event being emitted at all when sending credentials from the client to Sockethub.

The Sockethub logs also don't mention any server connection attempts:

```plain
  sockethub:core:wnJuJ9Ivg9qrjCn1AAAB socket.io connection +0ms
  sockethub:validate:wnJuJ9Ivg9qrjCn1AAAB applying schema validation for activity-object +0ms
  sockethub:validate:wnJuJ9Ivg9qrjCn1AAAB applying schema validation for credentials +0ms
  sockethub:core:wnJuJ9Ivg9qrjCn1AAAB credentials encrypted and saved +3s
```

silverbucket:
resolves #529


silverbucket:
@raucao this is ready for review!

silverbucket:
ping @raucao

raucao:
Just from a first quick glance, I think it would be nice to use clean callbacks directly where we can do them now, instead of still handing over to a generic event handler function and then figuring out what was coming in there.

raucao (approved):
Tested successfully by implementing it in Hyperchannel. Works both with a wrong network host, as well as when blocked on a VPN by Libera for example.

https://github.com/67P/hyperchannel/pull/274

silverbucket:
> Just from a first quick glance, I think that in the IRC example code, it would be nice to use clean callbacks directly where we can do them now, instead of still handing over to a generic event handler function and then figuring out what was coming in there.

Yeah I agree, I have had it in my head to refactor and simplify the examples further. I'll make a ticket it for it

PR #673, opened 2022-02-06
IRC SASL Support

### LINKED ISSUE #543: Support SASL Auth for IRC
Should be pretty straight-forward: https://github.com/ircanywhere/irc-factory/wiki/Setup-&-Usage

silverbucket:
Take over maintenance of [irc-socket](https://github.com/Havvy/irc-socket) published as [irc-socket-sasl](https://github.com/silverbucket/irc-socket). Which now contains SSL support and SASL authentication.

Update Sockethub IRC platform to make use of the new module.

Resolves #543


raucao:
Refs https://github.com/67P/hyperchannel/issues/240

raucao:
FYI: I'll wait with implementing/reviewing this until we have the connect decoupling merged in both SH and Hyperchannel, because it will make this much easier and also just nicer to use.

packages/sockethub-platform-irc/index.js
```diff
@@ -576,22 +578,27 @@ IRC.prototype.__connect = function (key, credentials, cb) {
   const is_secure = (typeof credentials.object.secure === 'boolean') ?
     credentials.object.secure : true;
 
-  const module_creds = {
+  const module_options = {
     username: credentials.object.nick,
```
raucao: I think if an `object.username` is set, it should be used instead of the nickname. Nicknames can be different from usernames.

silverbucket: Makes sense, added that as a possible property in the credentials.

raucao (approved):
dancing

PR #674, opened 2022-02-06
Switch to userAddress from username

### LINKED ISSUE #281: XMPP credentials: rename username to userAddress
It's actually a whole user address, so we should name it accordingly to prevent confusion in the platform.

silverbucket:
Resolves #281

### DIFF
--- a/packages/sockethub-platform-xmpp/src/utils.test.js
+++ b/packages/sockethub-platform-xmpp/src/utils.test.js
@@ -6,13 +6,23 @@ describe('Utils', () => {
   describe('buildXmppCredentials', () => {
     it('returns correct credential object used for xmpp.js connect', () => {
       expect(utils.buildXmppCredentials({object: {
-        userAddress: 'barney@dinosaur.com.au', port:123, server:'foo', password:'bar', resource: 'Home'
+        userAddress: 'barney@dinosaur.com.au', port:123, password:'bar', resource: 'Home'
       }})).to.eql({
         password: 'bar',
         service: "dinosaur.com.au",
         username: "barney",
         resource: 'Home'
       });
+    });
+  });
+  it('allows overriding server value', () => {
+    expect(utils.buildXmppCredentials({object: {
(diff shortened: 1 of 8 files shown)


silverbucket:
One question @raucao is with `util.BuildXMPPCredentials()` ... the existing logic has been to try to get the server name from the user address and if we get it, to override any existing `server` value sent in on the credentials. Is this correct? Wouldn't it be correct to not require `server`, however if it is specified, to keep that as the `server` value regardless of what's in the `userAddress`?

Example:
```
{
  server: "xmpp.foo.org",
  userAddress: "user@foo.org"
}
```

Is that a valid type of configuration? (A users XMPP server is different than their userAddress)

If not, then perhaps we shouldn't even allow the server property to be set as part of the credentials at all?

raucao:
> Is that a valid type of configuration? (A users XMPP server is different than their userAddress)

Yes, definitely valid! Especially important for connecting via Tor for example.

raucao (approved):
LGTM :+1:

silverbucket:
OK just updated to allow the `server` property to override the value got from the `userAddress` if it's supplied:
https://github.com/sockethub/sockethub/pull/674/commits/622d68ffbb5b4916d0f2b6bac43c989feb8c6402
Look good?

raucao:
> Look good?

Yes. :+1:

packages/sockethub-platform-xmpp/src/utils.test.js
```diff
@@ -15,4 +15,14 @@ describe('Utils', () => {
...
     });
   });
+  it('allows overriding server value', () => {
+    expect(utils.buildXmppCredentials({object: {
+        userAddress: 'barney@dinosaur.com.au', port:123, server:'foo', password:'bar', resource: 'Home'
+      }})).to.eql({
+      password: 'bar',
+      service: "foo",
+      username: "barney",
+      resource: 'Home'
```
raucao, line 25: The port is missing here.

PR #675, opened 2022-02-06
Rename AS observe to query

### LINKED ISSUE #647: Change AS verb "observe" to "query"
> I'm proposing to change it to "query" instead, because the term 1) makes more sense in the language itself and 2) is better suited for many more query types.

Background is adding query types for room info (#540) and other things that are not room attendance. And since the next release will break AS object notations, this would be a good time to introduce more breaking changes all at once to have avoid more breakage in the future.

silverbucket:
This is ready for review but is based on #671 which should be reviewed & merged first.

Resolves #647


packages/sockethub-platform-irc/index.js
```diff
@@ -417,9 +417,9 @@ IRC.prototype.update = function (job, credentials, done) {
 };
 
 /**
- * Function: observe
+ * Function: query
  *
- * Indicate an intent to observe something (ie. get a list of users in a room).
+ * Indicate an intent to query something (ie. get a list of users in a room).
```
raucao:
```suggestion
 * Indicate an intent to query something (e.g. get a list of users in a room).
```

raucao:
Using the old term "observe" results in the following Sockethub error message being sent to the client, instead of a validation error:

> TypeError: platform[jobData.msg.type] is not a function"

raucao:
One more thing I found: the XMPP platform is still sending the attendance list as type "observe" instead of "query".

silverbucket:
@raucao OK, this should be ready for review again. There weren't any up-front checks for platform method validation before being called, so this ended up touching a lot more files to get that in place. Fixed some tests too. Should be a lot more stable now.

raucao (approved):
Works great now!

PR #676, opened 2022-02-06
Fix XMPP Unit Tests

### LINKED ISSUE #658: XMPP data tests are not working
As per @raucao https://github.com/sockethub/sockethub/pull/653
I haven't had a chance to look into this, but didn't want to forget as no issue was created yet.

silverbucket:
This PR fixes the unit tests validation of XMPP platform methods. We now validate the calls to xmlFake rather than the send call. Caught a lot of bugs while doing this.

@raucao One thing I ran in to was that the `update()` method seemed to be setting the `presence` with wrong properties/values. I re-read the [RFC for `presence`](https://xmpp.org/rfcs/rfc3921.html#presence) and adjusted the method to what seems like a more compliant result. But can you double check this as well? I'm not 100% sure I got it right.

Resolves #658


packages/sockethub-platform-xmpp/src/index.js
```diff
@@ -313,7 +311,7 @@ class XMPP {
       const status = job.object.content || '';
       // setting presence
       this.debug('setting presence: ' + show + ' status: ' + status);
-      this.__client.send(xml(show, { type: status })).then(done);
+      this.__client.send(xml("presence", { show: show }, { status: status })).then(done);
```
raucao: If the variable name is the same as the property name, you only need to put the name in modern JS, like so:

```suggestion
      this.__client.send(xml("presence", { show }, { status })).then(done);
```

raucao:
> @raucao One thing I ran in to was that the update() method seemed to be setting the presence with wrong properties/values. I re-read the [RFC for presence](https://xmpp.org/rfcs/rfc3921.html#presence) and adjusted the method to what seems like a more compliant result. But can you double check this as well? I'm not 100% sure I got it right.

Could you explain what you thought was wrong, and what you think the right properties and values are? I'm a bit confused by the code there tbh.

silverbucket:
@raucao Well, this is how it had been implemented.

```javascript
const show = job.object.presence === 'available' ? 'chat' : '';
const status = job.object.content || '';
this.__client.send(xml(show, { type: status })).then(done);
```

So, if the job data was this:
```
{
  actor: actor,
  object: {
    type: 'presence',
    presence: 'available',
    status: 'available'
  }
}
```
The `xml()` function would be called with the following params `'chat', { type: '' }`. 

That didn't look right to me, so I checked the spec (linked above) and yeah, I was confused as to why it was like that in the first place. So I made changes in the PR and updated the tests, but I'm not sure if it's correct or not.

packages/sockethub-platform-xmpp/src/index.js
```diff
@@ -311,7 +311,7 @@ class XMPP {
       const status = job.object.content || '';
       // setting presence
       this.debug('setting presence: ' + show + ' status: ' + status);
-      this.__client.send(xml(show, { type: status })).then(done);
+      this.__client.send(xml("presence", { show }, { status })).then(done);
```
raucao: According to you explanation in https://github.com/sockethub/sockethub/pull/676#issuecomment-1030861554, doesn't this then never set presence for a `groupchat` as per spec? What does "show" actually mean here? (Sorry if the question is dumb, I'm still confused by this code and the spec at the moment.)

silverbucket: Looks like `show`, if set, can only be one of `dnd`, `away`, `chat`, `xa` ... if not set, client assumed to be available. https://xmpp.org/rfcs/rfc3921.html#stanzas-presence-children-show

So the code needs to be fixed more. I'm not sure what enum we should give the `presence` property in the long run (eg. do we just match the XMPP list?), but in the short run we can just say if it's `available` don't set the `show` property at all, and if it's anything else, set `away` ?

silverbucket: OK, I dug into it a bit further. I am still not 100% sure I understand how XMPP handles presence, because there's the presence `type` property and the `show` child element, both of which handle some sort of presence. https://xmpp.org/rfcs/rfc3921.html#stanzas-presence-children-show

I think I got a basic working version defined now, including cleaning up the presence object schema as it was slightly incorrect. Added some more tests, etc. Can you have a look, specifically at this commit, and let me know what you think? https://github.com/sockethub/sockethub/pull/676/commits/65eb793dfd74b9e823ac340edd497d3e31ce7b8a

silverbucket:
@raucao OK I've updated things so there are no properties or child elements used if there is not value set for them. This should be ready now.

raucao (approved):
I think this is good to merge now! :tada: 

How are we keeping track of breaking changes btw?

PR #677, opened 2022-02-06
use port as part of service

silverbucket:
Here's a fix based on https://github.com/xmppjs/xmpp.js/tree/main/packages/client#client-1


raucao (approved):
If that's how the XMPP library works, then it LGTM. :)

Issue #679, opened 2022-02-08
Actor type is wrong for room presence updates


raucao:
I think this should be "room"?

```json
{
  "type": "update",
  "actor": {
    "type": "person",
    "id": "kosmos@kosmos.chat"
  },
  "object": {
    "type": "presence",
    "presence": "online"
  },
  "target": {
    "id": "jimmy@kosmos.org/hyperchannel"
  },
  "context": "xmpp"
}
```

silverbucket:
Where did this object come from?

silverbucket:
This would be resolved by #540

PR #684, opened 2022-02-11
Unify schema validation

### LINKED ISSUE #681: Validate outgoing AS objects
Currently outgoing AS objects (those generated by irc2as specifically) are not being validated against the sockethub schemas. This means we can easily have inconsistencies with objects sent from sockethub to the client as we continue to improve and increase validation checks with object sent from the client to sockethub.

This was noticed when addressing https://github.com/sockethub/sockethub/issues/679 (irc2as generated AS object with an invalid actor type), and was the start of a larger code overhaul that I'm working on now.

silverbucket:
This turned in to quite a large PR, there are going to be breaking changes specifically with the format of some of the messages sent to the client during an IRC session, and also maybe some things related to presence.

* Move all schema validation logic and handling to `sockethub-schemas`
* `sockethub-schemas` exports helper functions for validating AS objects for use in tests or on the fly
* `irc2as` has unit tests to validate all of it's generated AS objects which are sent to the client
* `sockethub` hands off all validation to `sockethub-schemas`
* `sockethub-schemas` has near 100% test coverage
* `sockethub-schemas` has re-written logic for finding the best error message from the list of `ajv` errors when a schema fails validation. This is still not perfect, but much better than it was.

Resolves #681


raucao:
Wow, large indeed!

>  there are going to be breaking changes specifically with the format of some of the messages sent to the client during an IRC session, and also maybe some things related to presence

What exactly are the breaking changes? It takes much more time to adjust client code without knowing what needs adjusting.

packages/irc2as/src/index.test.js
```diff
@@ -0,0 +1,87 @@
...
+      return done(new Error('failed matching ' + JSON.stringify(stream)));
+    }
+    const err = schemas.validator.validateActivityStream(stream);
+    if (err) {
+      done(err + ' - ' + JSON.stringify(stream));
+    }
+  }
+}
+
+describe("irc2as converts incoming IRC data to ActivityStreams", () => {
```
raucao: This is the text that belongs in an "it" expectation. Conversely, the "it" text below is a "describe" text actually. If you actually use context and expectation blocks as intended, you will see that it also helps with better structuring specs, since "it" should mostly just contain the expectations (converts x to y), and multiple of them oftentimes, while "describe" should only describe *what* is being tested, not what the expectations are.

See here for a better explanation with example: https://stackoverflow.com/a/32055428

raucao (commented):
Just a comment about spec usage/wording for now. Have yet to test with Hyperchannel.

silverbucket:
You should be able to get some idea from checking out the irc2as diff, specifically as-emitter, i’m not at my computer until tomorrow.

packages/irc2as/src/index.test.js
silverbucket: Thanks for the tips, I'm always trying to get better at presenting unit tests. I tried my best to refactor these tests, however my hands are tied a bit as the tests for inputs/outputs really needs to run all in one test, due to the nature of the way irc2as works. One input doesn't equal one output, so they can't be broken into separate tests, and since irc2as buffers some lines of irc before generating an output, the memory state needs to be preserved for the whole series of inputs. Then at the end we make sure all outputs that we expect have been matched.

Still though, I think the structure of the tests are slightly better now.

silverbucket:
@raucao I've refactored the getErrorMessage() method to satisfy codeclimate, and made a best effort to improve the irc2as unit test structure. This is ready for review.

silverbucket:
To summarize some of the breaking changes with objects coming from `irc2as`:

* errors were sometimes represented as a `type` in the `object`, this has been changed to the `error` property (string) on the root of the AS object, as was settled on in sockethub core.
* actor and target have been swapped in the case that an action was an error. The actor is now the person who intiated it, rather than before where it was represented as the room (actor) sending an error to the user (target)
* context `irc` was added to all object
* `send` was changed  to `update` for errors specifcially.
* the topic object uses the standard `content` property for the topic, rather than a `topic` property.
* MOTD was changed to a long string thats appended per-line of irc input, rather than an array of lines one for each input.

That about sums up the breaking changes for incoming IRC AS to the client. There are likely a few other breaking things in here but hopefully they'll be very easy to identify (esp. considering the schema errors have been improved). This should be a major step towards tracking future breaking changes though, since we're validating everything upfront now.

silverbucket:
@raucao OK, I've added schema validation tests for output XMPP AS objects as well, to further tighten things up. So the objects should all be in a unified format now. The only remaining thing is with guessing the `type`... right now the guessing is `person` but we need some way of determining when to set type `room` instead. I could use your help on that as I don't have any ideas off the top of my head.

raucao:
There's no way of guessing type room from nothing but a JID string, because they look exactly the same as a person.

I guess the cheapest way to determine if a JID belongs to a room is to know beforehand which domains belong to a MUC service. For example when a `join` is received, it could remember the MUC domain as such.

Another way would be to actually discover the MUC domain in the process; something we need a message/function for anyway, to allow users to discover their own server's MUC. This could, and maybe should, be combined with discovering and caching the other information about the service in one go, because the MUC info is requested and received in a generic `disco#items` request:

https://xmpp.org/extensions/xep-0045.html#disco-service

silverbucket:
@raucao OK, I've updated the ticket to not resolve #679 ... barring that, does the PR work for you?

raucao (approved):
:+1:

PR #686, opened 2022-02-12
Clean up session callbacks properly

silverbucket:
This is a bugfix to remove callbacks when a session is terminated.

### DIFF
--- a/packages/sockethub/src/janitor.ts
+++ b/packages/sockethub/src/janitor.ts
@@ -1,7 +1,7 @@
 import debug from 'debug';
 
-import { platformInstances } from './platform-instance';
-import serve from "./serve";
+import PlatformInstance, { platformInstances } from './platform-instance';
+import serve, { SocketInstance } from "./serve";
 
 const rmLog = debug('sockethub:janitor');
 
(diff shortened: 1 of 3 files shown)


packages/sockethub/src/janitor.ts
```diff
@@ -57,6 +57,9 @@ function removeStaleSocketSessions(platformInstance, sockets) {
       rmLog('removing stale socket session reference ' + sessionId + ' in platform instance '
         + platformInstance.id);
       platformInstance.sessions.delete(sessionId);
+      for (let key in platformInstance.sessionCallbacks) {
+        platformInstance.sessionCallbacks[key].delete(sessionId);
+      }
```
raucao: I think in modern JS, you could write this like so:

```js
      for (const sessionCallback of platformInstance.sessionCallbacks) {
        sessionCallback.delete(sessionId);
      }
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

silverbucket: This isn't the case for objects, you can't iterate over the values of them without some additional code.

raucao: Ah, sorry. I thought it was an array for some reason.

silverbucket:
@raucao think I could get an approval on this one?

PR #691, opened 2022-02-25
Use ISO 8601 dates for "published" property of messages

### LINKED ISSUE #247: Format of published field changed
In the latest Sockethub release (3.1.2), the message's `published` field format changed from being an ISO 8601 datetime string (e.g. "2019-06-14T06:32:42.025Z") to a unix timestamp string (e.g. "1577981398164").

Was this on purpose or did it happen by accident? I couldn't find anything about it in the release notes or the documentation.

raucao:
fixes #247

### DIFF
--- a/packages/sockethub-schemas/src/object-types.js
+++ b/packages/sockethub-schemas/src/object-types.js
@@ -54,6 +54,9 @@ const objectTypes = {
       },
       "content": {
         "type": "string"
+      },
+      "published": {
+        "enum": ["this is not a timestamp"]
       }
     }
   },
--- a/packages/sockethub-platform-xmpp/src/incoming-handlers.js
+++ b/packages/sockethub-platform-xmpp/src/incoming-handlers.js
@@ -169,6 +169,8 @@ class IncomingHandlers {
     }
 
     if (timestamp) { activity.published = timestamp; }
+    // TODO Use once schema tests are working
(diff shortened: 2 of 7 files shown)


raucao:
@silverbucket I added the `published` property to the `message` schema, but it doesn't seem to actually be applied in the validations. As such, no validations for the message are trustworthy at the moment, so I pushed a commit to prove this in CI.

I have added the code that would actually make a real validation/test for the ISO date format work, and commented it out until CI is actually failing for the AS object violating the schema.

packages/sockethub-schemas/src/object-types.js
```diff
@@ -54,6 +54,9 @@ const objectTypes = {
       },
       "content": {
         "type": "string"
+      },
+      "published": {
+        "enum": ["this is not a timestamp"]
```
silverbucket, line 59: I'm a little confused as to why this is an enum of `this is not a timestamp` ?

raucao: https://github.com/sockethub/sockethub/pull/691#issuecomment-1051270745

It should test for the value only being valid when it's "this is not a timestamp", so since it is a timestamp, this proves in CI that the validations/tests are faulty.

silverbucket:
Resolved in https://github.com/sockethub/sockethub/pull/745

PR #704, opened 2022-03-05
Migrate to @sockethub namespace, use Nx

silverbucket:
* Move all packages to within the `@sockethub` namespace, except for the main `sockethub` package
* Switch much of the task commands to use Nx
* Consolidate eslint/tsconfig settings accross all packages, ability to ling both JS and TS


silverbucket:
@galfert @raucao  - This PR is ready for review. Sorry that it's so huge, it was like pulling a thread from a sweater. I created a secondary branch off this one, where I'm working on creating a sockethub meta package so that `npm i sockethub` will still work, and will come with all the platforms ready to run. So that means that this PR is fully working within the repo, but shouldn't be published as there will be no sockethub package published.

Most of the files here are renamed/package dependencies and linting configs, so don't be too intimidated. Very little actual code (almost none) changed.

raucao (approved):
thumbs-up-knight

PR #708, opened 2022-03-17
Fix dependencies in schemas

silverbucket:
Move devDependencies to dependencies, as now validation happens within the schemas module whereas before it was just providing the schema files.


raucao (approved):
LGTM

Issue #707, opened 2022-03-17
Improve linting


silverbucket:
Currently the following are disabled in the `tsconfig.base.json` which should be enabled:

```
    "strictNullChecks": false,
    "noImplicitAny": false,
    "noImplicitThis": false,
```

And in the `.eslintrc.json`
```
    "extends": [
//        "eslint:recommended",
...
//            "extends": [
//                "plugin:@typescript-eslint/recommended"
//            ]
```

PR #709, opened 2022-03-17
Sockethub as meta package

### LINKED ISSUE #706: Split sockethub package into sockethub and @sockethub/server
In order to have a `sockethub` parent package, while also being able to pick and choose from `@sockethub/server` and perhaps one or two platforms (`@sockethub/platform-xmpp` `@sockethub/platform-irc`). A number of things would need to happen:

* Platforms for the server would be peer dependencies
* Platforms for the sockethub package would be normal dependencies, as they are now
* The integration with @sockethub/client would need to be more flexible, currently the client js file is copied from the client repo into sockethub/dist
* The location of the sockethub bin file would need to be considered

For now, I won't publish a @sockethub/server package, and will keep `sockethub` as it is, and eventually we can address these issues (and likely others) as an incremental improvement in increased modularity

silverbucket:
Overall this PR decouples the modules further, you can install just the `@sockethub/server` package without the platforms (`--omit optional` I believe). Then installed the specific platforms you want, create a sockethub config (there's an example provided) to specify whichever platform you installed, and it'll use just that.

There is a `sockethub` package still, which statically depends on all the platforms, and no config is required, everything installs and works as it currently does.

List of everything:

* new `packages/sockethub` package which is a "meta" package, pre-configured to install all platforms and behave as the current package
* `packages/server` has all the server code, with optional dependencies for the platforms, so it's possible to install it without the platforms. 
* consolidated and fixed the config file parsing, as it's become more critical in how sockethub is bootstraped
* the platform threads are no longer dependent on re-parsing the config, since all they needed from it was the redis connect info (passed via env)
* dependencies of all packages adjusted to be less coupled
* Docker, build and tests improved to be less tightly coupled
* platform init is simplified
* bootstrap is simplified 
* sockethub executable functionality was moved to src/index.ts and the executable just includes it and runs.

Depends on #704 
Resolves #706


silverbucket:
ready for review @raucao @galfert - again sorry for the big PR but a lot of files needed to be touched and moved around, dependencies adjusted and code decoupled, in order to make each module more stand-alone. Also there's now a `sockethub` package and `@sockethub/server` package. See PR description for details.

packages/sockethub/bin/sockethub
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env node
+const SockethubServer = require('@sockethub/server');
+SockethubServer();
```
raucao: This looks a bit weird to me. Usually, a capitalized variable should be a class/constructor, not a function that starts a server.

packages/sockethub/package.json
```diff
@@ -0,0 +1,52 @@
+{
```
raucao, line 1: Did you intentionally switch to tabs for this one? Seems to be inconsistent with the rest.

test/xmpp.test.js
```diff
@@ -71,15 +71,22 @@ describe('integration page', () => {
       }, genHandler(done));
     }).timeout(3000);
 
+    it('`create activity-object', () => {
```
raucao: This seems like it should be a `describe` block.

tsconfig.base.json
```diff
@@ -16,10 +16,10 @@
     ],
     "esModuleInterop": true,
     "alwaysStrict": true,
-    "strictNullChecks": false, // TODO improve linting with this set to true
-    "noImplicitAny": false, // TODO improve linting with this set to true
```
raucao, line 20: Seems like a good idea to me to warn for implicit `any` types.

raucao (commented):
Nice one! :clap: 

Left a few comments.

tsconfig.base.json
silverbucket: Yeah this is the plan, but I didn't want to spam this PR with all those linting fixes. So will make a separate PR where I'm just tightening up linting.

raucao: I meant removing the TODO, since it's a good reminder...

silverbucket: Yeah, there was a linting issue with comments in a JSON file I think, I moved them to this issue instead: :) https://github.com/sockethub/sockethub/issues/707

packages/sockethub/bin/sockethub
silverbucket: 👍

packages/sockethub/package.json
silverbucket: Oops, not sure how that happened.

raucao (approved):
ship-it-squirrel

PR #710, opened 2022-03-19
Remove `?.` conditional accessor from JS code

silverbucket:
It was only being used in one place, and I didn't notice the incompatibility as I'm running node `16.5` locally. This way we can keep node `12` support. Eventually this will be TypeScript so it doesn't matter that we don't use this in the JS.


raucao (approved):
:+1:

raucao:
Some tests are still failing in CI...

silverbucket:
> Some tests are still failing in CI...

@raucao should be fixed now. and the sooner we get the other PRs in the easier it'll be to fix any other bugs we find without causing tons of merge conflicts.

PR #711, opened 2022-03-20
Update schemas package to TypeScript

silverbucket:
Schemas platform is now all in TypeScript, improved build process to the `dist` folder includes typing, schemas and json schemas for each import in TS or JS code.
Depends on PRs #709 and #704


packages/schemas/package.json
```diff
@@ -8,28 +8,46 @@
   "keywords": [
     "sockethub",
     "schema",
-    "activity",
-    "streams"
+    "activity streams",
+    "activity objects",
+    "platforms"
```
raucao: I don't think this keyword makes sense in a global npm context.

packages/schemas/src/index.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  it('validate correct platform schema', () => {
+    let err = validatePlatformSchema(testPlatformSchemaData);
+    expect(err).to.equal("");
+  });
+  it('validate incorrect platform schema', () => {
+    let err = validatePlatformSchema({foo:'bar'});
+    expect(err).to.eql('platform schema failed to validate:  must have required property \'name\'');
+  });
+  it('add platform schema', () => {
+    addPlatformSchema(testPlatformSchemaData.credentials, `test-platform/credentials`);
```
raucao: The expectation is missing here.

packages/schemas/src/index.test.ts
```diff
@@ -0,0 +1,68 @@
...
+import ActivityObjectSchema from './schemas/activity-object';
+import {validatePlatformSchema, addPlatformSchema, validateActivityObject,
+  validateActivityStream, validateCredentials} from './validator';
+import testCredentialsData from './index.test.data.credentials';
+import testActivityObjectsData from './index.test.data.objects';
+import testActivityStreamsData from './index.test.data.streams';
+import testPlatformSchemaData from './index.test.data.platform';
+
+describe('Platform schema validation', () => {
+  it('validate correct platform schema', () => {
```
raucao: If you try to write the description in a way that grammatically fits with the "it", it's easy to see how it's supposed to be used. This one for example could be written as:

```suggestion
  it('returns an empty error for a valid schema', () => {
```

The `it` block is a description of the behavior you expect from a function or feature. If there is more context needed, then you can wrap it in another `describe` block. (E.g. `describe("valid schema"` and then multiple `it` blocks for the various things you expect it to do. Not necessary in this case.)

The `describe` block can also use the actual function name, like e.g. `describe("validatePlatformSchema()"`, so that it's easy to see what's being tested in the output. This way, you can also see that "add platform schema" is actually context/describe information, and not describing what `addPlatformSchema()` is supposed to result in.

raucao (commented):
Code LGTM. Left a few comments for improvements.

Everything seems to work fine, except for leaving channels:

```plain
  sockethub:middleware:message Error: platform type leave not supported by xmpp platform. (types: credentials, connect, update, send, join, query, request-friend, remove-fri
end, make-friend)                                                                                                                                                            
  sockethub:middleware:message     at Array.<anonymous> (/home/basti/src/sockethub/sockethub/packages/server/dist/middleware/validate.js:46:33)  
```

packages/schemas/package.json
silverbucket: yeah, changed to `sockethub platforms`

packages/schemas/src/index.test.ts
silverbucket: Nice catch, added a `getPlatformSchema()` func, and compare result.

packages/schemas/src/index.test.ts
silverbucket: Thanks for pointing this out and the overview, it's easy to get careless or lazy about this, but is much nicer/more pleasing when it's organized more clearly.

silverbucket:
> Code LGTM. Left a few comments for improvements.
> 
> Everything seems to work fine, except for leaving channels:
> 
> ```
>   sockethub:middleware:message Error: platform type leave not supported by xmpp platform. (types: credentials, connect, update, send, join, query, request-friend, remove-fri
> end, make-friend)                                                                                                                                                            
>   sockethub:middleware:message     at Array.<anonymous> (/home/basti/src/sockethub/sockethub/packages/server/dist/middleware/validate.js:46:33)  
> ```

That's actually really cool that it fails "cleanly" with `leave` because the platform schema doesn't define it in it's schema. Because it shows that the work in tightening up the schema validation works (eg. it doesn't fail unexpectedly and it doesn't sloppily succeed when it wasn't defined in the schema).

silverbucket:
@raucao all your points have been addressed, ready for review again.

raucao (approved):
Perfect! :+1:

PR #737, opened 2022-05-18
Option to disable unknown prop warnings

### LINKED ISSUE #660: Remove invalid property warnings
as.js middleware is warning about invalid properties, probably best to remove that entirely.

silverbucket:
Resolves #660

### DIFF
--- a/packages/activity-streams.js/src/activity-streams.js
+++ b/packages/activity-streams.js/src/activity-streams.js
@@ -28,10 +28,10 @@ const ee = EventEmitter(),
           'id', 'type', 'context',
           'alias', 'attachedTo', 'attachment', 'attributedTo', 'attributedWith',
           'content', 'contentMap', 'context', 'contextOf', 'name', 'endTime', 'generator',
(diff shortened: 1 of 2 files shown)


packages/activity-streams.js/src/activity-streams.js
```diff
@@ -61,6 +61,7 @@ const ee = EventEmitter(),
 
 let objs = new Map(),
     failOnUnknownObjectProperties = false,
+    warnOnUnknownObjectProperties = true,
```
raucao, line 64: Didn't we say it should not print to the console by default, since sending invalid data is a client issue and the client should receive that warning anyway?

silverbucket: When the server initializes the object, it disables the warnings here https://github.com/sockethub/sockethub/pull/737/files#diff-e1a9a5a584326eccba5e95ec2316a1484a3958a2eb01aa9aa8bf999eb7390baa

raucao (approved):
thumbs-up-asian-guy-gif

PR #738, opened 2022-05-18
Enable recommended linting

silverbucket:


silverbucket:
ping @raucao or @galfert for review

PR #745, opened 2022-06-01
Use ISO 8601 dates for incoming XMPP messages

### LINKED ISSUE #247: Format of published field changed

silverbucket:
Set the `published` property to `ISO 8601`.
Resolves #247

### DIFF
--- a/.github/workflows/compliance.yml
+++ b/.github/workflows/compliance.yml
@@ -5,7 +5,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [16.x]
+        node-version: [ 16 ]
     steps:
       - uses: actions/checkout@v2
       - name: Use Node.js ${{ matrix.node-version }}
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [ 14.x, 16.x ]
+        node-version: [ 14, 16 ]
     steps:
       - uses: actions/checkout@v2
       - name: Use Node.js ${{ matrix.node-version }}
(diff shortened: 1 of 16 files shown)


raucao:
What about #691? I already implemented this, and the tests should be red in a commit, before reviewing the functionality.

silverbucket:
@raucao this is ready for review, based on your previous PR and added schema support for the `date-time` format.

raucao:
My previous PR shows a testing bug where the schema isn't validated in the tests, which hasn't been fixed in that branch. You can push to the same branch to prove that the tests are actually validating the schema correctly, but it should be a different branch to fix that I think.

raucao:
> I'm sorry but I don't understand exactly what you are saying should be done, or what's wrong with this PR.

I'm saying there's no way for me to know that the validation is actually being applied now, other than jumping into the code myself. A PR that fixes the validations being applied would have a test that shows that they are applied. Same if it would just be a commit that uses my intentionally faulty schema.

It's just that since I had everything prepared in the other PR for exactly that, and the PR is still open, it wasn't obvious that you actually fixed the validations. And it's still not clear to me, because there's no commit message stating that, or reference to the PR that proved they were broken.

I'll just jump into the code and verify it manually be editing it again then.

raucao:
I checked, and the same bug for the tests still persists. They are green, despite the format not matching the schema.

I have pushed a similar commit to this branch to demonstrate the bug.

silverbucket:
@raucao Sorry for taking so long to get to this, summer break :) .. I finally figured out what was so confusing about this issue. The `published` field was being defined on an `ActivityObject` and no the base `ActivityStream`. This should be working 100% now.

silverbucket:
@raucao spoke too soon, coverage seems to be failing on github, though it's passing locally.

silverbucket:
@raucao OK, test added to schemas to bring coverage back up to 95%

raucao (approved):
:+1: :rocket:

PR #764, opened 2022-08-14
Use `postinstall` for `lerna bootstrap`

silverbucket:
* Use `postinstall` for `lerna bootstrap`
* Update github actions and docker
* Update docs for a more conventional quick start

### DIFF
--- a/.github/workflows/compliance.yml
+++ b/.github/workflows/compliance.yml
@@ -13,7 +13,7 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - run: npm install -g yarn
-      - run: yarn deps
+      - run: yarn install
       - run: yarn lint
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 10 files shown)


raucao (commented):
I tried a fresh checkout, and cannot `yarn run dev` immediately after `yarn install` and `yarn run build && yarn test` (both green). It's missing a default config file it seems:

> Error: Config file not found: sockethub.config.json

silverbucket:
@raucao It seems to have have something to do with the build not fully executing (even though it's a dependency for the run directive), I added another build step to the `postinstall` directive, with caching disabled. It seems to be working for me, can you see how it goes for you?

silverbucket:
eg. this command succeeded for me: 
`git clone git@github.com:sockethub/sockethub.git && cd sockethub && git co td/quick-start && yarn install && yarn run dev`

raucao:
Now it works, but it feels kind of wrong to build everything in the install command, which is only supposed to pull dependencies. If you say this should happen in the `dev` command, then wouldn't it be better to do it there?

raucao (approved):
:+1:

PR #766, opened 2022-08-22
Fix XMPP example - target format

silverbucket:
As a quick short-hand, if you use `an@xmpp.address: a message` in the text input, it uses the data left of the `:` as the target, this fixes the object format.


silverbucket:
ping @raucao for review

raucao:
Sorry, but I'm not familiar with the example code there at all. I think you should just merge example updates without review, when nobody jumps in after a couple of days.

PR #767, opened 2022-08-22
Remove the listeners for the stale socket session

silverbucket:
The `platform-instance` class was not having stale socket listener functions removed during termination.


silverbucket:
ping @raucao for review

raucao:
Could you explain what this caused/fixes? I do see session-related warnings in the console a lot, but no idea how this relates exactly.

silverbucket:
> Could you explain what this caused/fixes? I do see session-related warnings in the console a lot, but no idea how this relates exactly.

@raucao  This fixes the situation that a socket becomes stale (disconnect from client) however a handler for incoming messages for that socket is still registered. So the handler is fired, a warning that it's unable to find the socket is shown, and that's it. 

The original fix https://github.com/sockethub/sockethub/pull/767/commits/b6ad381f21d4623d667898bda33b6a80e2440841 simply makes sure to remove the listener as we already keep a record of its function signature.

However, I decided to add coverage to this PR to test for this (since it would help answer your question) but testing the janitor file was problematic, I re-wrote it as a Singleton class which made testing much easier, we now test for this and all (or most of) the other conditions of cleanup/removal that the janitor attends to. 

The downside is now the PR is much bigger and the fix is hidden within all the refactored code.

raucao:
I think I saw this before testing this PR, but when I just closed my Hyperchannel tab, and then re-opened it, the XMPP connection fails with:

```
[xmpp] failed to connect to XMPP server:  
Object { type: "connect", context: "xmpp", actor: {…}, error: "record not found for key: jimmy@kosmos.org/hyperchannel" }
​
actor: Object { id: "jimmy@kosmos.org/hyperchannel", type: "person", name: "jimmy" }
context: "xmpp"
error: "record not found for key: jimmy@kosmos.org/hyperchannel"
type: "connect"
```

silverbucket:
@raucao I have been working for the past 3 days on splitting the code out into different pacakges, and improving the testing. When I try this on that branch (which is branched from this branch) the issue seems to be resolved for me. Can you try it there and see how it behaves for you? https://github.com/sockethub/sockethub/pull/769

I had been having some issues updating to the new version of the redis library (which dependent modules I maintain, `redis-connection-pool` and `secure-store-redis` had been updates to use), and deciding decoupling things more would help reduce, or at least isolate, bugs when some piece needs to change.

raucao:
> Can you try it there and see how it behaves for you?

Just tried, and it works all great on that branch!

silverbucket:
> Just tried, and it works all great on that branch!

Awesome, in that case I suggest we go ahead and merge this (just the listener bugfix). That larger PR https://github.com/sockethub/sockethub/pull/769 is branched from this one, so it'll become the next to review to merge to master. As well, after that then Feeds branch https://github.com/sockethub/sockethub/pull/771 which is branched from that one will become the next to review to merge to master.

raucao (approved):
thumbs-up-asian-guy-gif

PR #769, opened 2022-08-27
Split out data-layer and crypto to own packages

silverbucket:
This PR splits out the crypto and data-layer code into their own packages, for improved separation of concerns, API management, isolation and testing.

* Crypto
Fairly simple module, needed to be split out to avoid circular dependencies from other packages.

* Data Layer
This package handles any communication to/from redis, to remove the dependency on the server package and make testing new versions of redis libraries easier. It consists of two classes, CredentialsStore and JobQueue. 

Along the way several issues were resolved regarding failed jobs, they should now always be resolved (callback called) even if they were queued behind a failing job that needed to be successful (eg. a connect, then a join). Previously the join could never get resolved when the connect fails (and the platform destroyed).

- [x] - ensure there are no unneeded dependencies still defined in `crypto`, `data-layer` or `server` packages


silverbucket:
@raucao ok, cleaned up a few things, this huge PR is ready for review

raucao (approved):
I glanced over the code changes, all LGTM. Good job with the tests and wording/clarity of spec descriptions and examples btw.

Also ran this with Hyperchannel with both IRC and XMPP connections and it all works well.

silverbucket:
Cheers, thanks!

PR #771, opened 2022-08-28
Fix Feeds results, with integration tests

silverbucket:
This makes sure the feed fetching results, which is an array of activity objects, is passed back to the client. Includes integration tests to make sure any breakage is caught in the future.


test/integration.test.js
```diff
@@ -74,6 +74,27 @@ describe('Page', () => {
       }
     });
 
+    describe('Feeds', () => {
+      it(`fetch an XML feed`, (done) => {
```
raucao: it fetch? :)

The "it" blocks are for expectations ("it does xyz"). "describe" is for descriptions.

test/integration.test.js
```diff
@@ -74,6 +74,27 @@ describe('Page', () => {
...
+        }, (msg) => {
+          console.log(`Feed fetch callback! `, msg);
+          expect(msg.length).to.eql(20);
+          done();
+        });
+      });
+    })
+
     describe('XMPP', () => {
       it('sends credentials', (done) => {
```
raucao: This is missing a describe block for what exactly is expected to send credentials.

packages/server/src/platform-instance.ts
```diff
@@ -177,29 +177,29 @@ export default class PlatformInstance {
...
+      await this.sendToClient(job.sessionId, payload);
     }
 
     // let all related peers know of result as an independent message
     // (not as part of a job completion, or failure)
-    await this.broadcastToSharedPeers(job.sessionId, msg);
+    await this.broadcastToSharedPeers(job.sessionId, payload);
 
-    if ((this.config.persist) && (this.config.requireCredentials.includes(job.msg.type))) {
+    if ((this.config.persist) && (this.config.requireCredentials.includes(payload.type))) {
```
raucao:
```suggestion
    if (this.config.persist && this.config.requireCredentials.includes(payload.type)) {
```

raucao (commented):
Seems to work fine in general! :+1: 

Some nitpick comments about spec wording.

Aside from that, I think there's a bug with the object content. `brief_html` and `brief_text` contain the entire content from the source article, while `html` contains the abbreviated content. Shouldn't that be the other way around?

silverbucket:
@raucao this is ready for review, you already did the heavy lifting in your first review, just need a last check and approval.

packages/platform-feeds/index.js
```diff
@@ -257,10 +246,10 @@ function buildFeedObject(dateNum, item) {
     date: item.date,
     datenum: dateNum,
     tags: item.categories,
-    text: item.summary,
-    html: item.summary,
-    brief_text: item.description,
-    brief_html: item.description,
+    brief_text: item.summary,
+    brief_html: item.summary,
```
raucao: Why are these two different keys, when they both use the same data? The `brief_text` contains HTML in all the feeds I tried with the example. It's usually also the same as the full HTML (i.e. not brief), but I don't remember which feed I used before where I got different data in brief and full (but in reverse).

packages/platform-feeds/index.js
```diff
@@ -257,10 +246,10 @@ function buildFeedObject(dateNum, item) {
...
+    text: item.description,
+    html: item.description,
```
raucao: Same as above: why have a `text` key here that duplicates the (HTML) data from the other key?

raucao (commented):
Left a couple more comments about the properties.

There's one more thing I just saw: there's a `date` and a `datenum` property in the AS `object` for when the feed item was published, but no `published` property for the activity itself. Shouldn't this data move to `published` to better conform with AS conventions? (Or at least be copied there?)

raucao:
... btw, we could also merge as is and address the functional questions in new PRs later.

silverbucket:
@raucao hey, for the `html` vs `text` fields, what do you think about stripping the html from the text fields? If the source text is not html, they'll both be plain text, but if the source is html, then there will be the text alternative.

As for the date, what do you suggest? that there be `published` property at the root of each activity stream (eg. one property for each AS in the array) `feed[2].published`. I guess that would be the most uniform as that's how we do it for XMPP yeah? I'm just wondering if we want to differentiate between the date of the post, vs the date the post was fetched, not sure.

raucao:
> @raucao hey, for the `html` vs `text` fields, what do you think about stripping the html from the text fields? If the source text is not html, they'll both be plain text, but if the source is html, then there will be the text alternative.

I just had a quick look at the [RSS](https://www.rssboard.org/rss-specification) and [ATOM](https://validator.w3.org/feed/docs/atom.html) specs. In RSS, encoded HTML is *allowed* in `description`, but it's up to the client to parse. In ATOM, text elements like `summary` and `content` should have a `type` attribute that specifies if it's HTML or plaint text, or even something else.

I'm not sure if the Sockethub platform is the right place to implement content formats and conversions to be honest. But it should always send all possible information to the client, of course. It would certainly make client implementation more convenient if at least the content type information would be there for every item, but I'm not sure doubling or even quadrupling the content payload is the way to go.

So I would maybe propose to only send a `summary`/brief value, if it's actually different from the `description` or `content` (i.e. the feed is an ATOM feed and content is different from summary), and I would probably not send `_text` and `_html`, but just `content` plus `content_type` if it's available.

> As for the date, what do you suggest? that there be `published` property at the root of each activity stream (eg. one property for each AS in the array) `feed[2].published`. I guess that would be the most uniform as that's how we do it for XMPP yeah? I'm just wondering if we want to differentiate between the date of the post, vs the date the post was fetched, not sure.

The client knows when the post was fetched, but there are two possible dates from the feed that it may want: the date when the item was originally published (I believe that's what `published` is for in AS objects), and the date that the item was last updated. The latter is actually the only required date element for ATOM feed entries, while `published` is optional, and `pubDate` is optional in RSS, too.

silverbucket:
@raucao OK I think I've addressed both of your comments regarding duplicate payload (i now also provide contentType) and the date property naming to published. Let me know what you think.

test/integration.test.js
```diff
@@ -41,84 +41,151 @@ describe('Page', () => {
...
+            // expect(msg.length).to.eql(20);
+            // for (const m of msg) {
+            //   expect(m.object.content.length).to.be.true;
+            //   expect(m.object.type).to.equal('feedEntry');
+            //   expect(m.object.contentType).to.equal('html');
+            //   expect(m.actor.type).to.equal('feed');
+            //   expect(m.type).to.equal('post');
+            // }
+            // done();
+          // });
```
raucao: Looks like this shouldn't be commented...

silverbucket: Yeah, in the process of trying to figure out what broke the integration test, but otherwise the PR is complete. (I won't merge until integration tests are working again)

silverbucket:
Just FYI - the compliance is all passing, it's just that I consolidated the jobs `test`, `lint`, `coverage` into one `compliance` to reduce noise. the last time these ran was before `compliance` existed so it thinks they are missing.

raucao (changes requested):
I'm sorry, but the [first feed I tried to fetch](https://kosmos.social/@raucao.rss) failed with the following error:

    "TypeError: Cannot read properties of null (reading 'trim')"

raucao:
I'm wondering what `status: true` is:

Screenshot from 2022-10-12 12-05-30

raucao:
One more thing: I don't think there should be a value for "updated" when the source feed does not contain that property.

silverbucket:
> One more thing: I don't think there should be a value for `updated` when the source feed does not contain that property.

The `feedparser` API doesn't really distinguish, the `date` property is always either the same as `pubDate` or more recent (e.g. updated). So I've added a comparison, if they are the same it's set to undefined (and therefore not passed along).

raucao (approved):
Looking all good now! :clap: :rocket:

PR #779, opened 2022-10-10
Improve client libraries build process

silverbucket:
Using rollup instead of webpack, include both `esm` and `commonjs` import/require paths, as well as browser bundles. Inspired by the way it's done with `socket.io-client`, applied to both `activity-streams.js` and `sockethub-client`.

* Compile typescript to both `esm` and `commonjs` 
* Use `rollup` instead of `webpack` to build browser-includable files
* Update `package.json` with various entry points for different outputs, esm, cjs, types, etc.


silverbucket:
@raucao This PR is essentially just related to the build process for the JS clients (activity streams and sockethub client), I output more options so they can be included via `require`, `import` or as a browser bundle. In the case of activity streams, I converted the source to typescript, however didn't change any of the code logic at all. 

So it shouldn't be much to review if you have a few minutes.

silverbucket:
ping @raucao

raucao:
> I output more options so they can be included via `require`, `import` or as a browser bundle.

I couldn't find any documentation updates about this in the changeset? At least not in a `README` or `.md` document.

silverbucket:
> > I output more options so they can be included via `require`, `import` or as a browser bundle.
> 
> I couldn't find any documentation updates about this in the changeset? At least not in a `README` or `.md` document.

Check the package.json, there are more types of importa defined pointing to various build files, it behaves better being used as an esm or commonjs module, points to the types file, as well as providing dist bundles for browser script includes

raucao:
Cool. But shouldn't all ways of using the library be documented for developers? Or how would they know about those browser bundles for example?

silverbucket:
@raucao Somehow I never noticed that the client library didn't have a readme? Anyway, added the basics to both the as and client libraries. I plan to do much more documentation on this later. Currently working on a totally redone set of working examples to replace the old ones as well.

raucao (changes requested):
I quickly smoke-tested this with Hyperchannel and unfortunately it's broken as is. There are a bunch of replay messages for AS objects, but connections to both IRC and XMPP don't succeed. But there is also no error message.

@silverbucket Could you run this with Hyperchannel and see how it works for you?

silverbucket:
@raucao Odd, yup I'm getting the save behavior on `hyperchannel` that you described, -- the XMPP connect works however at first glance it seems the joins are completing but the app never gets the callbacks. However it works as expected with the sockethub XMPP example, so there's some deviation in usage. I'll investigate further when I have a bit more time.

silverbucket:
@raucao all tests passing, and the "dont pass same object back on completion" change was reverted, so this should be good to go. Verified it works on Hyperchannel.

raucao (approved):
:+1:

PR #782, opened 2022-10-17
Rewritten Examples as own module

silverbucket:
Creation of `@sockethub/examples` module, written in Svelte + Tailwind CSS. 
- [x] Set up `examples` package and framework
- [x] Changes to `server` package to handle serving from the  `examples` package. Removed old `views` folder
- [x] Dummy Example
- [x] Feeds Example
- [x] IRC Example
- [x] XMPP Example
- [x] Handle incoming messages
- [x] Only show logs for current platform
- [x] Responsive design
- [x] Replace any CSS with Tailwind


packages/crypto/yarn.lock
```diff
@@ -289,9 +289,15 @@ acorn-walk@^8.1.1:
   integrity sha512-k+iyHEuPgSw6SbuDpGQM+06HQUa04DZ3o+F6CSzXMvvI5KMvnaEqXe+YVe555R9nn6GPt404fos4wcgpw12SDA==
 
 acorn@^8.4.1, acorn@^8.8.0:
+<<<<<<< HEAD
```
raucao: There are various git merge conflicts committed in the pushed lockfiles...

silverbucket: yeah I noticed too, I thought I had regenerated them and apparently it didn't actually do it, so I went through and manually did it afterward.

packages/examples/src/app.css
```diff
@@ -21,6 +21,14 @@
...
         white-space: pre-wrap;
     }
+    #incomingMessages { list-style-type: none; margin: 0; padding: 0; overflow: scroll; height: 20em;}
+    #incomingMessages li { padding: 0.1em 1%; }
+    #incomingMessages li:nth-child(odd) { background: #eee; }
+    #incomingMessagesContainer {
+        height: 20rem;
+        display: block;
+        border: 1px solid #ccc;
+    }
```
raucao: Any reasons for not using Tailwind for these?

silverbucket: I just copied some of this over from the old examples and then adjusted it to work, but good point - I will replace it with tailwind

silverbucket:
@raucao or @galfert I think this PR is finally ready for review. There are some test entries for old node versions which don't run anymore, so those can be ignored. Improving styling, CSS, tooltips, etc. can be done in another ticket as this one is fairly huge already.

packages/examples/src/routes/feeds/+page.svelte
```diff
@@ -0,0 +1,72 @@
...
+
+  function sendFetch() {
+    send(getASObj("fetch"))
+  }
+</script>
+
+
+<Intro title="Feeds Platform Example">
+  <title>Feeds Example</title>
+  <p>The feeds platform fetches takes an RSS/ATOM feed URL and parses it, returning an array of Activity Objects for each entry.</p>
```
raucao:
```suggestion
  <p>The feeds platform takes an RSS/ATOM feed URL, fetches and parses it, and returns an array of Activity Objects for each entry.</p>
```

raucao (commented):
I tried out the examples, and mostly just found the UI to be very raw and making it a bit difficult to figure out. But as you say, that should be fixed in different PRs.

Just a couple comments left:

* Some of the files are formatted with tabs, while most are not (which I would prefer and is the case with the rest of the codebase). There were also formatting changes (e.g. single quotes to double quotes), which mix in a lot of unrelated changes with the functionality in this PR.
* The examples test fails for me with `tests/test.ts:3:1 › index page has expected h1`. There is also a message about a missing browser executable from Playwright, which wasn't fixed with the proposed command (`npx playwright install`).
* Not sure if the test failure is related, but it would be nice to have the examples linked from the root URL, when they are enabled. I vaguely remember this having been the case before. The root URL returns a 404 now.

silverbucket:
Alright, finally got around to addressing the last of the issues, fixing the linting and tests. 

> * Not sure if the test failure is related, but it would be nice to have the examples linked from the root URL, when they are enabled. I vaguely remember this having been the case before. The root URL returns a 404 now.

It was always only available from `/examples` so I've left it like that for now.

silverbucket:
@raucao this is finally done, again! A lot of changes in this PR, but another big step in the right direction as I slowly improve the integration testing process, as well as add more checks and lining to improve proactive catching of issues.

PR #824, opened 2023-04-02
Switch to pnpm

silverbucket:
This PR removes the use of `yarn`, in exchange for `pnpm`, it removes `lerna` and simplifies the use of `nx`. 
`35k` LOCs removed just from `yarn.lock` files! 🥳


silverbucket:
@raucao I'm going to force push this, as it should greatly minimize the renovate notifications, and git rid of all the spammy PRs that are open right now.

raucao:
Cool.

BTW, I refuse to remember which repo prefers which package manager nowadays. If you're like me, you can use [ni](https://github.com/antfu/ni) in every repo and always automatically run the right install command. Then you see what it's using on install and can remember to use the correct command for scripts.

PR #829, opened 2023-05-03
1. Improve linting (examples and related areas)

silverbucket:
This is one step toward improved linting, focusing on some issues with the examples package, and a more unified compliance process to make sure we are checking each package uniformly.

There will be more of these coming, trying to keep each PR somewhat focused though, and introduce more strict checking little by little.


silverbucket:
@raucao solved the playwright test failure, this PR is ready for review.

PR #833, opened 2023-05-08
5. Prettier

silverbucket:


silverbucket:
> I can't say I like the indentation, because literally everything I've written and maintained over the last decade was 2 spaces. That said, whatever works for you is best. Merge at will please.

Yeah, I used to agree completely, but after using Prettier on a few projects for the past few years I have to admit that I really like not just the 4 spaces, but also the uniformity of code in general, switching between things, etc. It's nice to have sensible formatting decisions that span projects.

I had to update some files and this branch would be too busy to resolve the merge conflicts, so I created a second branch, with the same changes. So I'll delete this PR and force-merge the other.

Issue #563, opened 2021-08-31
Investigate scaling/re-using Redis connections for Bull


silverbucket:
This implementation was used however as @Cobertos noticed, wasn't operating correctly with custom Redis config. More time will need to go into investigating this implementation, and making sure we're managing the Redis connections properly, with tests and verification of connections used/cleaned up.

- https://github.com/OptimalBits/bull/blob/master/PATTERNS.md#reusing-redis-connections
- https://github.com/OptimalBits/bull/issues/2081
- https://github.com/OptimalBits/bull/issues/1798

Not high on the priority list, unless we start running into Redis connection issues.

raucao:
> Not high on the priority list, unless we start running into Redis connection issues.

I agree. I'd even say it's "low", not "medium".

silverbucket:
This is no longer applicable as we switched to `bullmq` the rewrite of `bull`

PR #836, opened 2023-05-10
Rewrite to support newer version of Redis, with checks & integration tests

### LINKED ISSUE #690: Alert ahead of time when Redis isn't listening
Currently, it seems Sockethub wont fail until someone tries to send it an AS to handle. It should verify redis is listening and the connection is working as part of init.

silverbucket:
* Uses the latest `secure-store-redis` which I rewrote to no longer use `generic-pool` behind the scenes as it was causing issues with the newer `redis` client.
* Fail immediately if we don't connect to Redis on startup.
* Refactored the code in `@sockethub/data-layer` to adapt to the newer, simplified, `SecureStoreRedis` API
* Refactored code that make sure of the `@sockethub/data-layer` lib in the `@sockethub/server` package.
* Added integration tests for simple writing to/from Redis using the `@sockethub/data-layer`. Checks both `CredentialsStore`, `JobQueue` and `JobWorker` (bullmq)
* Improved the `@sockethub/examples` package.
* Fixed some issues with `@sockethub/data-layer/CredentialsStore`, to fail clearly when credentials are not found or cannot be verified.
* Redis config now only uses URLs: `"url": "redis://127.0.0.1:6379"`

Resolves #690 #563


packages/crypto/src/index.ts
```diff
@@ -31,6 +31,7 @@ export class Crypto {
     }
 
     encrypt(json: IActivityStream, secret: string): string {
+        Crypto.ensureSecret(secret);
```
raucao, line 34: Why not just `this.ensureSecret`? Makes it clear that it's part of the same class.

raucao (approved):
Code LGTM. Only had one nitpick.

Successfully tested locally with node.js 18 and Hyperchannel.

packages/crypto/src/index.ts
silverbucket: It's because the `ensureSecret` function is static, so you can't use `this` annotation as it is not instance-specific.

PR #837, opened 2023-05-14
Ensure bootstrap initialization is only run once

silverbucket:
There was a bug (that didn't seem to produce any symptoms) where the bootstrap process was being run twice during startup. This fixes the issue and adds some tests.


raucao (approved):
LGTM

PR #839, opened 2023-06-25
Simplified examples with editable actor

silverbucket:
Simplified some of the code, allow for setting the actor. You can't edit the text-area JSON, which is just for displaying what the JSON objects look like.


silverbucket:
@raucao all tests passing

PR #872, opened 2024-08-23
Documentation: Fix broken links

silverbucket:


raucao (approved):

raucao:
My review wasn't good enough. The feeds source link goes to `dist` instead of `src`, which isn't checked into the repo.

silverbucket:
@raucao doh! here's the fix https://github.com/sockethub/sockethub/pull/873

PR #882, opened 2025-01-19
XMPP presence "room" vs "person"

### LINKED ISSUE #679: Actor type is wrong for room presence updates

silverbucket:
If the incoming `from` has a slash, we assume that's a person, otherwise it's a room. 
Resolves #679


raucao:
This is a not a safe way of identifying the nature of a JID. A person is not required to announce a resource (the part after the slash).

If there's no surrounding information that clarifies the nature, then you'd have to do service discovery on the domain. Otherwise you will inevitably end up with false positives.

silverbucket:
@raucao OK I'll do some more digging and see what kind of data we get back from an `update` event, maybe there's something more helpful there.

PR #886, opened 2025-02-01
Migrate to Bun

silverbucket:
This PR migrates to `bun` which was less of a task than migrating to `deno`, but doesn't rule out migrating to `deno` in the future (and in fact will help make that task less painful).

- Removes ~10k+ of boilerplate bundle and build configs, `rollup`, `tsconfig`
- Most packages don't need to build at all anymore (just client-side libs, `client` and `activity-streams`)
- Removes `nx` dependency
- Removes deprecated `karma` in place of `web-test-runner`
- Fixes/improves tests
- Significant performance improvements (~10x)


silverbucket:
@raucao @galfert @gregkare - This PR migrates from node.js to bun, which is a relatively painless drop-in replacement for node.js, while supporting execution of `ts` files, greatly improved performance. Ready for review.

(Also- it seems the kosmos kredits tag checker is down?)

raucao:
> (Also- it seems the kosmos kredits tag checker is down?)

I can see it in the list:

Screenshot from 2025-02-10 15-27-36

raucao:
image

silverbucket:
> With a fresh checkout, the install and linting worked fine, but running dev fails like this:

Ah, there is a necessary build step to make the client libraries and a static setup of resources for the examples. I've added it as part of the `dev` step now so it should all work with just the `install` and `dev`.

raucao:
OK. Does that build continuously on changes, like `dev` did before?

silverbucket:
> OK. Does that build/reload continuously on changes, like `dev` did before?

No live-reloading, but sockethub never really had that (at least not for many many years) just due to the complexity of the builds, I never bothered and always ctrl-c'd and restarted to test changes.

I can revist live reloading though, as things are a lot simpler now, but I'd like to keep that out of scope of this PR. Created a ticket for it: https://github.com/sockethub/sockethub/issues/890

> Another thing that I'm missing, which was pretty useful, is the coloring of the log lines (different colors for different modules).

Just fixed this, I was using the --filter param to run the sockethub server from root, but that output strips colorization. So we now cd to the dir and run from there.

raucao (approved):

silverbucket:
> I pushed a small change, moving the build step to `postinstall`, so that it's all there after setting up the repo (and for that 1 failing test also), but not being rebuilt on every dev restart (which is kinda slow compared to just running the server code). Hope you like it that way, too?

Even better, thanks!

PR #887, opened 2025-02-02
Upgrade to latest Svelte/Sveltekit and eslint

silverbucket:
Update deps, fix breaking changes.


silverbucket:
@raucao @galfert or @galfert  - can I get a quick check here? no functional changes, I just needed to update some major versions and redo their configs.

PR #897, opened 2025-02-20
Migrate examples app to Svelte 5

silverbucket:
This PR handles all the migration details of the examples app from Svelte 3 to Svelte 5.


raucao:
I don't really know anything about Svelte, but I also don't think the examples need in-depth reviews at this point.

PR #898, opened 2025-02-21
Hot reload using `bun run dev`

### LINKED ISSUE #890: Live-reloading dev runs
See if we can use live-reloading for dev runs of sockethub on changes.

silverbucket:
The live reloading is built-in to bun, so this was a lot easier than expected.
Resolves #890

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -54,8 +54,8 @@
     "socket.io": "4.7.5"
   },
   "scripts": {
-    "start": "DEBUG=sockethub* ./bin/sockethub",
-    "dev": "DEBUG=sockethub* ./bin/sockethub --examples"
+    "start": "DEBUG=sockethub* bun run ./bin/sockethub",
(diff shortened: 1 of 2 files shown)


packages/server/package.json
```diff
@@ -54,8 +54,8 @@
     "socket.io": "4.7.5"
   },
   "scripts": {
-    "start": "DEBUG=sockethub* ./bin/sockethub",
-    "dev": "DEBUG=sockethub* ./bin/sockethub --examples"
+    "start": "DEBUG=sockethub* bun run ./bin/sockethub",
+    "dev": "--hot DEBUG=sockethub* bun run ./bin/sockethub --examples"
```
raucao: Shouldn't it be `bun run --hot`?

silverbucket: nice catch, i'm blind:D fixed

raucao (approved):
You can squash commits when merging this one...

PR #899, opened 2025-02-22
Add Sentry support

### LINKED ISSUE #817: Sentry integration
We just set up Sentry for Kosmos, and I think it would be nice to be able to report Sockethub exceptions to it (optionally). Would be useful for anyone running Sockethub in production and wanting to know and be alerted about errors/exceptions that might occur.

https://docs.sentry.io/platforms/node/

silverbucket:
Added basic sentry support. Only is enabled if a Sentry DNS is supplied (either with a `--sentry.dsn [dsn]` arg, or in the config file.
Resolves #817

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ node_modules
 **/dist/
 .vscode
 .DS_Store
+sockethub.config.json
--- a/bun.lock
+++ b/bun.lock
@@ -2739,9 +2739,27 @@
 
     "@sinonjs/commons/type-detect": ["type-detect@4.0.8", "", {}, "sha512-0fr/mIH1dlO+x7TlcMy+bIDqKPsw/70tVyeHW787goQjhmqaZe10uwLujubK9q9Lg6Fiho1KUKDYz0Z7k7g5/g=="],
 
+    "@sockethub/activity-streams/@types/bun": ["@types/bun@1.2.3", "", { "dependencies": { "bun-types": "1.2.3" } }, "sha512-054h79ipETRfjtsCW9qJK8Ipof67Pw9bodFWmkfkaUaRiIQ1dIV2VTlheshlBx3mpKr0KeK8VqnMMCtgN9rQtw=="],
+
+    "@sockethub/client/@types/bun": ["@types/bun@1.2.3", "", { "dependencies": { "bun-types": "1.2.3" } }, "sha512-054h79ipETRfjtsCW9qJK8Ipof67Pw9bodFWmkfkaUaRiIQ1dIV2VTlheshlBx3mpKr0KeK8VqnMMCtgN9rQtw=="],
+
+    "@sockethub/crypto/@types/bun": ["@types/bun@1.2.3", "", { "dependencies": { "bun-types": "1.2.3" } }, "sha512-054h79ipETRfjtsCW9qJK8Ipof67Pw9bodFWmkfkaUaRiIQ1dIV2VTlheshlBx3mpKr0KeK8VqnMMCtgN9rQtw=="],
+
+    "@sockethub/data-layer/@types/bun": ["@types/bun@1.2.3", "", { "dependencies": { "bun-types": "1.2.3" } }, "sha512-054h79ipETRfjtsCW9qJK8Ipof67Pw9bodFWmkfkaUaRiIQ1dIV2VTlheshlBx3mpKr0KeK8VqnMMCtgN9rQtw=="],
+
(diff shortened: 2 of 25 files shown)


silverbucket:
An easy way to test this would be to add the `sentry` line to your existing config, or if you use defaults, then create a `sockethub.config.json` in the directory where you run sockethub from (or if it's run from the `bun run dev` then it should go in `packages/sockethub/` and add the dsn to the sentry prop:
```json
{
  ...
  "sentry": "https://sentrydsn"
  ...
}
```

raucao (commented):
Usually, the config key for a Sentry DSN includes the "DSN" part, because there may be other Sentry-related configs as well.

silverbucket:
> Usually, the config key for a Sentry DSN includes the "DSN" part, because there may be other Sentry-related configs as well.

Good point, I've expanded the sentry config as an object, rather than a string. 

I decided to just have the entire sentry config object be treated as if it was the same as the real sentry config. So, if `sentry.dsn` is set, we pass sentry the entire `sentry` config object.

silverbucket:
@raucao forgot to commit my final changes before that last comment :D it's pushed now.

raucao (changes requested):
The command line option doesn't work anymore, so I guess that part from the PR description has changed?

Also, there is a console log statement somewhere which seems to have been left in accidentally:

```
 DEBUG=sockethub* ./bin/sockethub --examples --sentry "http://855118badcc3c6e864281fdc963b0095@sentry.kosmos.local/4"                                                            
  sockethub:server:bootstrap:config initializing config +0ms                                                                                                                      
[                                                                                                                                                                                 
  {                                                                                                                                                                               
    $ref: "#/definitions/type/feed",                                                                                                                                              
  }, {                                                                                                                                                                            
    $ref: "#/definitions/type/message",                                                                                                                                           
  }, {                                                                                                                                                                            
    $ref: "#/definitions/type/me",                                                                                                                                                
  }, {                                                                                                                                                                            
    $ref: "#/definitions/type/person",                                                                                                                                            
  }, {                                                                                                                                                                            
    $ref: "#/definitions/type/room",                                                                                                                                              
  }, {                                                                                                                                                                            
    $ref: "#/definitions/type/service",                                                                                                                                           
  }, {                                                                                                                                                                            
    $ref: "#/definitions/type/website",                                                                                                                                           
  }, {                                                                                                                                                                            
    $ref: "#/definitions/type/attendance",                                                                                                                                        
  }, {                                                                                                                                                                            
    $ref: "#/definitions/type/presence",                                                                                                                                          
  }, {                                                                                                                                                                            
    $ref: "#/definitions/type/relationship",                                                                                                                                      
  }, {                                                                                                                                                                            
    $ref: "#/definitions/type/topic",
  }, {
    $ref: "#/definitions/type/address",
  }
]
```

raucao:
I'm not sure where or when something changed there, but in Hyperchannel, running this branch throws a `Failed to load Sockethub libs: TypeError: window.SockethubClient is not a constructor`.

The server is running and the client JS file is served and retrieved the same as before.

silverbucket:
> Also, there is a console log statement somewhere which seems to have been left in accidentally

This has been there for a while actually, it's showing the loading of the schemas, but yeah I've removed it as it should be a log statement or nothing.

raucao:
The CLI flag is working now, but the client still throws the same error. It seems to have initialized an object in the global context before, and now it doesn't.

silverbucket:
> The CLI flag is working now, but the client still throws the same error. It seems to have initialized an object in the global context before, and now it doesn't.
> 
> (I run `install` every time I fetch new changes, so that it should build all necessary files before I test.)
> 
> Edit: FYI, I don't see the client being fetched from the instance the same way in the examples.

Can confirm, seeing this as well. Not sure when it happened - as it likely wasn't this PR - it's likely the way it's built for the browser, somehow the object isn't being set to the window. I'll look into it.

silverbucket:
@raucao I made some adjustments to the way the client libraries are generated, but in general the browser libraries that are built need the `type="module"` in the script tag. I've also made a change to the integration tests that should catch this in the future.
Submitted this small hyperchannel fix as well: https://github.com/67P/hyperchannel/pull/299

silverbucket:
@raucao OK, we now properly catch both `uncaughtException` and `unhandledRejection` events, as well as in a few other key places. Most of the codebase should be covered now, the core sockethub init and handling of incoming and outgoing websocket messages, and the platform thread which runs the platform and sends in jobs, returns the results to the queue.

raucao (approved):
Confirmed to work with my exception now:

Screenshot from 2025-02-26 11-50-09

applause

raucao:
@silverbucket Can you disconnect that Sentry project from GitHub, please? It's not necessary to be connected just to receive errors.

silverbucket:
@raucao yeah, sorry, I thought I did but forgot. I have removed the github integration.

PR #902, opened 2025-02-24
New Platform: Metadata

silverbucket:
Given a URL, fetches the resource looking for Open Graph data (currently all that's supported).

Request object:
```json
{
  "context": "metadata",
  "type": "fetch",
  "actor": {
    "id": "https://sockethub.org",
    "type": "website"
  }
}
```

Response object:
```json
{
  "context": "metadata",
  "type": "fetch",
  "actor": {
    "id": "https://sockethub.org",
    "type": "website"
  },
  "object": {
    "type": "page",
    ...
  }
}
```


raucao:
There's official AS2 vocab for a web page btw:

https://www.w3.org/TR/activitystreams-vocabulary/#dfn-page

silverbucket:
@raucao yeah I shared that same link in the ticket discussing the feature. Also in the description of this PR I'm using the `page` type, but just for the object. The reason being we had already been using `website` for the actor/target already.
 
I can change it everywhere though and we can just go with `page` for the actor. 

The content is also of type page, but contains a lot more data that the basic props defined in the page spec. The `image` array is also from the spec (example 82)... I wonder if there any examples of a real actual web page being represented aside from simple a name and URL. 

I was also looking in to ActivityPub the other day. A lot more definitions for social media type stuff, but nothing really stood out, but if something looks interesting to you please share.

silverbucket:
ping @raucao

raucao:
```
{
  "context": "metadata",
  "type": "fetch",
  "actor": {
    "id": "https://sockethub.org",
    "type": "website"
  },
  "id": "1"
}```

Shouldn't the `actor` be the `target` there instead?

**Comment by silverbucket** [comment]
# Sockethub - ActivityStreams Structure

> Shouldn't the `actor` be a `target` there instead (since the actor doing the fetch/request action is the user/client)?

Use to be, yeah, but actually I'd like to change it, as the more I think about it in this context it is confusing and should be simplified in some manner. (I mention it in the description of this PR, the second bullet point, but will elaborate here).  

## Current behavior
Currently (this PR not withstanding) we need to always have an actor of some sort, this is due to previous design decisions but could probably be lifted if decided that is desired.
So, for non-persistent platforms (platforms which don't involved any type of auth) we currently have an object that looks like:
*request*
```json
{
  "context": "metadata",
  "type": "fetch",
  "actor": {
    "id": "aDummyID"
  },
  "target": {
    "id": "https://sockethub.org",
    "type": "website"
  }
}
```

The actor means nothing really. Some entity that is not tied to an account, at this point there is not such thing as an account for both the metadata and feeds platforms. This "person" wants to fetch a target which is the website. 
However, the return value would look something like this:
*response*
```json
{
  "context": "metadata",
  "type": "fetch",
  "actor": {
    "id": "aDummyID"
  },
 "object": {
    "type": "page",
    "language": "en",
    "title": "Sockethub - a polyglot messaging service",
    "name": "Sockethub",
    "description": "The polyglot solution to the federated social web",
    "image": [
      {
        "url": "/res/img/sockethub-logo.png",
        "type": "png"
      }
    ],
    "url": "https://sockethub.org",
    "favicon": "/favicon.ico",
    "charset": "ISO-8859-1"
  }
  "target": {
    "id": "https://sockethub.org",
    "type": "website"
  },
  "id": "1"
}
```

This was how we agreed it should be a long time ago, but over this seems to make less and less sense and adds some unnecessary noise and thought to look at and decide you can ignore the actor. And then looking at the target, and thinking, "Oh, this is the site we're fetching".
Alternatively, the actor and target could be switched, but again it is essentially describing unnecessary meta data (e.g. "aDummyID" was the person that requested this fetch).
This makes less sense when you think about possible caching of the results and sending them to multiple parties (sharing a link in a chat room). 
## Proposals
If we're just talking about what the end result should look like, I would propose two possible solutions.
### Actor is Website
The person who fetches the document doesn't really matter in this case. When actors matter, is when they are posting to a page, or sending a message etc. 
We could make the argument that the actor is the website. That's what this PR does . So essentially, when a request is made, we're starting the AS which is then filled out by Sockethub:
*request*
```json
{
  "context": "metadata",
  "type": "fetch",
  "actor": {
    "id": "https://sockethub.org",
    "type": "website"
  }
}
```

This would be "filled out" by Sockethub:
*response*
```json
{
  "context": "metadata",
  "type": "fetch",
  "actor": {
    "id": "https://sockethub.org",
    "type": "website"
  },
 "object": {
    "type": "page",
    "language": "en",
    "title": "Sockethub - a polyglot messaging service",
    "name": "Sockethub",
    "description": "The polyglot solution to the federated social web",
    "image": [
      {
        "url": "/res/img/sockethub-logo.png",
        "type": "png"
      }
    ],
    "url": "https://sockethub.org",
    "favicon": "/favicon.ico",
    "charset": "ISO-8859-1"
  },
  "id": "1"
}
```

In this case, the response does not modify the original object, just fills out the `object` property. That's because typically Sockethub is not an actor or a target in and of itself, so it doesn't make sense that Sockethub would "respond" by flipping the actor and target. 
Think of when a message is sent on the XMPP platform, the actor sends a message to target, Sockethub confirms this message by sending back the *same* object that the actor sent, just with the request ID, and a potential error property, indicating that it was processed and either successfully sent or not (but not whether a response was received).
It's still a little odd as the "actor" isn't doing anything at all, it's just a website. 
### No Actor at all
The other object would be to remove the actor entirely, and instead you would "start" the object that you want to be filled out. So you would send a message like:
*request*
```json
{
  "context": "metadata",
  "type": "fetch",
  "object": {
    "id": "https://sockethub.org",
    "type": "page"
  }
}
```

Which is the starting point for what the website might look like. In fact, this could, on its own, be a descriptive enough object - it just has a URL and nothing else (no metadata at all).
But generally the response would be completed and look like:
```json
{
  "context": "metadata",
  "type": "fetch",
  "object": {
    "type": "page",
    "language": "en",
    "title": "Sockethub - a polyglot messaging service",
    "name": "Sockethub",
    "description": "The polyglot solution to the federated social web",
    "image": [
      {
        "url": "/res/img/sockethub-logo.png",
        "type": "png"
      }
    ],
    "url": "https://sockethub.org",
    "favicon": "/favicon.ico",
    "charset": "ISO-8859-1"
  },
  "id": "1"
}
```
So both the request and response could essentially be "processed" without any further information, it's just that Sockethub filled out the second object better. 
Think of a front-end web app where a user posts a link, the same function that displays the completed request (favicon, description, etc) could also be the one that shows the placeholder link at first (it would just be an object, of type page, that has a link). 
#### Downside
The downside here would be there would likely be a little more work done to make *no* assumptions that an actor property ever has to exist. So it would take a bit more work.


Took me a while to get a chance to write this out. Hope it's clear, let me know what you think. I kind of like the idea of having no actor at all but didn't want to make bigger changes without first having this exact discussion.

raucao:
The actor that actually fetches the metadata would be Sockethub in this case, no?

To be honest, it seems to me like ActivityStreams might just not be good fit for all of the communication between SH clients and servers. I don't see valid `@context` properties for AS in the examples anyway. Perhaps it makes more sense to use a custom context and schema for stuff like this, and only use AS when there's a clear, and ideally widely used and interoperable, AS mapping/format for data that is sent or received?

silverbucket:
> The actor that actually fetches the metadata would be Sockethub in this case, no?

That was what we had agreed on before, but as I mentioned above in "Current Behavior" this seems to make less and less sense as I've worked on sockethub, which is why I made the other proposals. 

> To be honest, it seems to me like ActivityStreams might just not be good fit for all of the communication between SH clients and servers. I don't see valid `@context` properties for AS in the examples anyway. Perhaps it makes more sense to use a custom context and schema for stuff like this, and only use AS when there's a clear, and ideally widely used and interoperable, AS mapping/format for data that is sent or received?

This has always been the plan, but even with custom contexts, we're still going to use the AS format. However, i think that's aside from the point. 

The point here is that, Sockethub is an AS "handler", that does one of the following:

1. Filling in relevant data to a bare-bones AS object or
2. Delivering an AS object to it's target

Then it makes less sense to generate AS objects which get flipped by Sockethub and sent back. Rather than an AS object that's like "here's what I know about this website, e.g. just a URL" and Sockethub fills it out with any other data it can find and sends it back.

I think over time this seems more and more correct to me, and actually simplifies the reasoning. Thinking of Sockethub as an active participant is confusing, rather than just thinking of it as a facilitator. 

"Send me your AS objects and I'll try to handle them", instead of "send me a command and I'll return a response".

PR #905, opened 2025-03-01
Add disconnect() function for XMPP and IRC platforms

silverbucket:
Added `disctonnect()` functions for both IRC and XMPP platforms, which calls cleanup and sets `initialized` to `false`, which - when checked with the `janitor` process - will indicate the thread can be destroyed.

### DIFF
--- a/packages/platform-irc/src/index.ts
+++ b/packages/platform-irc/src/index.ts
@@ -502,6 +502,22 @@ export default class IRC implements PlatformInterface {
         });
     }
 
+    /**
+     * Disconnect IRC client
+     * @param {object} job activity streams object
+     * @param done
+     *
+     * @example
+     *
+     * {
+     *    context: 'irc',
+     *    type: 'disconnect',
+     *    actor: {
+     *      id: 'slvrbckt@irc.freenode.net',
+     *      type: 'person'
+     *    }
+     *  }
+     */
     disconnect(job: ActivityStream, done: PlatformCallback) {
         this.debug(`disconnect called for ${job.actor.id}`);
         this.client(done);
--- a/packages/platform-irc/src/index.ts
+++ b/packages/platform-irc/src/index.ts
@@ -502,15 +502,20 @@ export default class IRC implements PlatformInterface {
         });
     }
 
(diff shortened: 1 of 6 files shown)


silverbucket:
The stop() function is already what is called, you don't see it?

As for the presence, perhaps that could be done in disconnect() which then calls cleanup for the stop() call. 

Or we could leave it for now and not set a mandatory presence for disconnect.

silverbucket:
@raucao https://github.com/sockethub/sockethub/pull/905/files#diff-239ab23866265dc277975c3836ddf851a836a4304efd0f75c18ce527da3d4701R550

raucao:
Yes, I missed that line. Thanks.

I'm not sure if we should always send the presence update or not. Technically, the spec says it should be done on every clean disconnect which is the idea of calling `stop()` in the cleanup. Let's make up our mind about this later.

packages/platform-irc/src/index.ts
```diff
@@ -502,15 +502,36 @@ export default class IRC implements PlatformInterface {
...
+     *    type: 'disconnect',
+     *    actor: {
+     *      id: 'slvrbckt@irc.freenode.net',
+     *      type: 'person'
+     *    }
+     *  }
+     */
+    disconnect(job: ActivityStream, done: PlatformCallback) {
+        this.debug(`disconnect called for ${job.actor.id}`);
+        this.client(done);
```
raucao: I think this is supposed to call `cleanup`, not `client`, no?

silverbucket:
> Let's make up our mind about this later.

OK should we merge this now and add the status update if we feel the need?

raucao:
> OK should we merge this now and add the status update if we feel the need?

Yes?

silverbucket:
@raucao OK I just need your approval then :)

raucao:
Oops, there was a comment that wasn't published...

silverbucket:
@raucao nice catch, I guess I should make sure I have some tests there before I merge.

silverbucket:
@raucao Tests added for both IRC and XMPP disconnect and cleanup. Ready for a final review

raucao (approved):
thumbs-up-asian-guy-gif

PR #927, opened 2025-10-15
Improve documentation

silverbucket:


docs/client/using-platforms.md
```diff
@@ -0,0 +1,486 @@
...
+        }
+    });
+});
+```
+
+### Credential Security
+
+- Never log or expose credentials in client-side code
+- Use environment variables for server-side credential storage
+- Implement proper authentication flows for production apps
```
raucao, line 456: It looks like these have been generated, because they don't really fit for Sockethub. Operators do not have to store credentials at all, and Sockethub handles the encryption automatically, no?

Also, what does "implement proper authentication flows for production apps" mean for Sockethub?

silverbucket: Yeah a lot of these docs were migrated from the old sockethub wiki (which I closed) and then updated by Claude, guess he went a little overboard here. 

I think it's mostly a useless set of statements that have no bearing. I'll delete them. I'm going to strip a lot of this down.

raucao: OK, cool.

Just a thing about merging with kredits labels but no review (which also results in no review kredits for things caught after merge): I think it's not the idea of kredits to earn them for committing unnecessary things and then earning more for removing said things. So maybe either wait for a full review when there's a kredits label, esp. medium or large. (For small ones it's usually fine not to wait). In other Kosmos repos a `kredits-2` is a usually a fairly large contribution, like a whole feature or similar size (a few full hours of work on average).

silverbucket: Yup, this did actually already take a few hours of work at least, which is why I gave it a `kosmos-2` aside from unexpected bugs, I don't usually attached a second kosmos label to PRs when I just need to correct my own mistake. 

I've always gone by your rule of thumb which is that, most things are `kosmos-2`, 3 should already be several days of hard work, and 1 should be little things here and there than were not very complicated (maybe done in 30-1hr or less).

As for the reviews, we spoke about it maybe 6-9 months ago? I had several PRs which were growing very stale after many months of no review. You mentioned that I don't need to wait for a review on every PR. So I went ahead and merged them. 

The PRs up right now are specifically for you to review, so I won't merge them before that, however for this documentation I just wanted to get the main hauling and consolidation of work done. The problem was just that the AI embelished too much here, I'm going to trim it down, I'm just busy with another PR right now. When I do, I won't mark another kosmos label to that PR.

raucao: Great, completely agree and we're on the same page.

I'm looking forward to having AI help with creating new platforms based on everything being documented properly like this!

PR #935, opened 2025-10-19
Reconnect test and more integration coverage

silverbucket:
This PR addresses an important bug related to credential handling during reconnection and adds comprehensive integration tests. The core fix prevents platform instances from crashing when receiving invalid credentials after initialization.

* Added reconnection tests (browser refresh, using correct and incorrect credentials)
* Improved multiclient tests (correct creds or wrong creds)
* Fixed up basic tests
* Improved test coverage
* Applied some fixes to sockethub related to handling bad creds on an existing platform instance (do not bring down instance, but reject job).

### DIFF
--- a/packages/server/src/platform.ts
+++ b/packages/server/src/platform.ts
@@ -187,7 +187,8 @@ function getJobHandler(): JobHandler {
                 // yet initialized, because they need to authenticate before they are initialized.
 
                 // Get credentialsHash for validation.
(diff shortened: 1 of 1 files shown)


raucao (approved):
This seems to work like a charm. I wasn't able to crash Sockethub anymore, and all refreshes worked without any problems. In fact, it now also replays channel history and attendance in Hyperchannel, which wasn't the case before.

thumbs-up-barney-yes

PR #1028, opened 2026-02-06
fix(platform-irc): Prevent unauthenticated IRC session sharing

### LINKED ISSUE #545: Prevent different unauthenticated sessions from using the same username
Currently, you can just use the same username as any other user on the same Sockethub instance, without the IRC server being able to discern between the two (or more).

silverbucket:
## Summary

This PR fixes persistent-session sharing for username-only IRC credentials.

Before: if two users connected with the same `context + actor.id` and no password,
Sockethub could treat those credentials as shareable and attach both sockets to one
platform thread.

Now:
- Session-share checks are owned by the data layer (`CredentialsStore.get`).
- Share attempts with a non-empty password remain allowed.
- Share attempts without a non-empty password are rejected with:
  - `username already in use`

This keeps anonymous/username-only sessions isolated per socket while preserving
normal authenticated sharing behavior.

## Reconnect Behavior During Janitor Grace

A narrow reconnect exception is supported for anonymous credentials:
- Prior session must be stale (disconnected socket, not yet janitor-cleaned).
- Reconnecting request must come from the same client IP.

If both conditions match, reconnect is allowed.

## Configuration

Added/used credential-check config options:
- `credentialCheck.reconnectIpSource`: `socket` (default) or `proxy`
- `credentialCheck.proxyHeader`: header name when using proxy mode (default `x-forwarded-for`)

## Errors

Different failure cases now stay distinct:
- Credential mismatch/validation failures: `invalid credentials for <actor>`
- Username-only share conflict: `username already in use`

## Documentation

This behavior is documented in:
- `docs/configuration.md`
- `docs/client-guide.md`
- `packages/server/README.md`

Resolves https://github.com/sockethub/sockethub/issues/545
Follow-up hardening ticket (actor churn abuse): https://github.com/sockethub/sockethub/issues/1037


raucao (changes requested):
I used this branch and was able to write messages as the same username on Libera from two different browser sessions at the same time. So the original issue persists.

packages/data-layer/integration/redis.integration.ts
```diff
@@ -22,9 +22,10 @@ const creds: CredentialsObject = {
     },
     object: {
         type: "credentials",
+        password: "secret",
     },
 };
-const credsHash = "e591ec978a505aee278f372354c229d165d2c096";
+const credsHash = crypto.objectHash(creds.object);
```
silverbucket: This should be static, not dynamically generate, to ensure no changes break the hashing.

packages/server/src/middleware/credential-check.ts
```diff
@@ -38,7 +38,7 @@ export default function credentialCheck(
         };
 
         credentialsStore
-            .get(msg.actor.id)
+            .get(msg.actor.id, undefined, { requirePassword: true })
```
silverbucket: No need for an additional param, this should be the default and only behavior.

packages/data-layer/src/credentials-store.ts
```diff
@@ -56,11 +56,16 @@ export async function resetSharedCredentialsRedisConnection(): Promise<void> {
 export interface CredentialsStoreInterface {
     get(
         actor: string,
-        credentialsHash: string | undefined,
+        credentialsHash?: string,
+        options?: CredentialsValidationOptions,
```
silverbucket, line 60: No need

silverbucket:
@raucao the issue is that Hyperchannel still expects any value in the callback (for credentials) to be an error, whereas now, sockethub is standardized on returning the object, which could contain an error property or not.  This is already the case for normal messages.
https://github.com/67P/hyperchannel/pull/304

silverbucket:
@raucao can I get your approval for this one? i think we sorted everything out, the fix is in a hyperchannel PR. I plan to make more improvements to hyperchannel once I get several of these larger PRs merged. Nearing 5.0 release!

raucao (approved):

PR #1038, opened 2026-02-20
feat!: canonical extended AS2 @context routing and validation

silverbucket:
## Summary
This PR standardizes Sockethub around one canonical AS payload shape and one canonical way to choose platform routing/validation: platform context URL in `@context`.

It also makes `@sockethub/client` the primary way to generate `@context` so app code does not need to hand-build context arrays.

### 1) Canonical message shape
Messages now use canonical fields:
- `@context` (array)
- `type`
- `actor`
- optional `object`, `target`, `error`, etc.

Legacy message forms are removed.

---

### 2) Sockethub client: first-class context generation
Preferred API:
- `sc.contextFor(platform)`
- `SockethubClient.contextFor(platform)`

Example:
```js
const msg = {
  '@context': sc.contextFor('xmpp'),
  type: 'connect',
  actor: { id: 'jimmy@prosody/SockethubTest', type: 'person' }
};
```

Generated JSON:
```json
{
  "@context": [
    "https://www.w3.org/ns/activitystreams",
    "https://sockethub.org/ns/context/v1.jsonld",
    "https://sockethub.org/ns/context/platform/xmpp/v1.jsonld"
  ],
  "type": "connect",
  "actor": {
    "id": "jimmy@prosody/SockethubTest",
    "type": "person"
  }
}
```

---

### 3) `@context` format and meaning (with example)
Example `@context`:
```json
[
  "https://www.w3.org/ns/activitystreams",
  "https://sockethub.org/ns/context/v1.jsonld",
  "https://sockethub.org/ns/context/platform/xmpp/v1.jsonld"
]
```

What each entry means:
1. `https://www.w3.org/ns/activitystreams`
   Base AS2 vocabulary.
2. `https://sockethub.org/ns/context/v1.jsonld`
   Sockethub base vocabulary.
3. `https://sockethub.org/ns/context/platform/xmpp/v1.jsonld`
   Platform identity for routing + schema lookup.

Server resolves this third URL to platform ID (`xmpp`) and validates against that platform’s schema.

---

### 4) Actor/target validation now uses field-specific type lists
Validation now reports allowed types by field (`actor` vs `target`) instead of one large mixed list.

Example bad input:
```json
{
  "@context": [
    "https://www.w3.org/ns/activitystreams",
    "https://sockethub.org/ns/context/v1.jsonld",
    "https://sockethub.org/ns/context/platform/fakeplatform/v1.jsonld"
  ],
  "type": "join",
  "actor": "irc://uuu@localhost",
  "target": "irc://irc.example.net/a-room"
}
```

Example error (target):
```text
[target]: must match exactly one schema in oneOf: person, room, service, feed, website, address
```

This makes error output match actual expected role types.

---

### 5) `packages/schemas` now generates canonical publishable artifacts
Generated artifacts include:
- platform JSON-LD contexts
- platform message schemas
- platform credentials schemas
- context lookup maps

Example generated map snippet:
```json
{
  "contextToPlatformId": {
    "https://sockethub.org/ns/context/platform/xmpp/v1.jsonld": "xmpp"
  },
  "contextToSchemaId": {
    "https://sockethub.org/ns/context/platform/xmpp/v1.jsonld": "https://sockethub.org/schemas/as2/platform/xmpp/v1/messages.schema.json"
  },
  "contextToCredentialsSchemaId": {
    "https://sockethub.org/ns/context/platform/xmpp/v1.jsonld": "https://sockethub.org/schemas/as2/platform/xmpp/v1/credentials.schema.json"
  }
}
```

---

## Breaking Changes
- Legacy/non-canonical message formats are rejected.
- Routing by old `context` field is removed.
- Canonical `@context` with exactly one registered platform context URL is required.

---

## Schema Publishing Proposal (`website-sockethub.org`)
Use `website-sockethub.org` as the publication target for generated contexts/schemas.

### Build timing
- On every merge to `master` in `sockethub`, build `packages/schemas` artifacts.
- Upload build artifacts from CI.

### Build verification
Before publish, verify:
1. all JSON files parse
2. map entries point to files that exist
3. each registered platform context has matching message + credentials schemas

### Publish timing
- Alpha phase: publish on every green `master` merge.

### Publish flow
1. `sockethub` CI builds artifacts and stores them as workflow artifact.
2. `website-sockethub.org` workflow pulls artifact by source commit SHA.
3. Copy files into website public tree:
   - `/ns/context/...`
   - `/schemas/as2/...`
4. Commit with source SHA in commit message.
5. Deploy website.

### Publish rules
- Only publish from green CI.
- Keep published versioned files immutable.
- Never delete old published versions.


packages/activity-streams/src/activity-streams.ts
```diff
@@ -87,14 +86,7 @@ const baseProps = {
...
 } as const;
-const rename: Record<string, string> = {
-    "@id": "id",
-    "@type": "type",
-    verb: "type",
-    displayName: "name",
-    objectType: "type",
-    platform: "context",
-};
+const rename: Record<string, string> = {};
```
silverbucket: Maybe just remove rename functionality altogether.

packages/irc2as/src/as-emitter.js
```diff
@@ -16,7 +16,11 @@ export class ASEmitter {
 
     __generalError(nick, content) {
         return {
-            context: "irc",
+            "@context": [
+                "https://www.w3.org/ns/activitystreams",
+                "https://sockethub.org/ns/context/v1.jsonld",
+                "https://sockethub.org/ns/context/platform/irc/v1.jsonld",
```
silverbucket: get the contexts from the user of this library rather than hard coding them.

packages/platform-dummy/src/index.ts
```diff
@@ -29,6 +29,12 @@ export default class Dummy implements PlatformInterface {
         return {
             name: "dummy",
             version: packageJSON.version,
+            as2: {
```
silverbucket: maybe as part of the existing schema property? or named something related, as as2 is too specific.

packages/schemas/src/generated/contexts/platform/dummy/v1.jsonld
silverbucket, line 1: Should schemas have hardcoded info for each platform? Or should this be defined in the platforms themselves, with the universal info injected from schemas package.

packages/activity-streams/src/activity-streams.ts
```diff
@@ -180,12 +169,6 @@ function validateObject(
 
     for (const key of unknownKeys) {
         const ao = incomingObj as ActivityObject;
-        if (rename[key]) {
-            // rename property instead of fail
-            renameProp(ao as Record<string, unknown>, key);
```
silverbucket, line 193: Make sure if the rename functionality is mentioned in the docs its removed.

packages/irc2as/src/as-emitter.js
```diff
@@ -1,10 +1,18 @@
 const EVENT_INCOMING = "incoming";
 const EVENT_ERROR = "error";
+export const DEFAULT_CONTEXTS = [
```
silverbucket: if something changes in sockethub schemas, or in the irc platform, these defaults become wrong, I'd rather they are not in this package at all and injected during instantiation. it could be required if necessary.

packages/schemas/scripts/export-canonical-assets.ts
```diff
@@ -48,16 +69,102 @@ mkdirSync(srcGeneratedContextsDir, { recursive: true });
...
 
-const platformSchemas: Array<PlatformSchemaShape> = [
-    (await import("../../platform-dummy/src/index.ts")).default.prototype
-        .schema,
-    (await import("../../platform-feeds/src/schema.ts")).default,
-    (await import("../../platform-irc/src/schema.ts")).PlatformIrcSchema,
-    (await import("../../platform-metadata/src/schema.ts"))
-        .PlatformMetadataSchema,
-    (await import("../../platform-xmpp/src/schema.js")).PlatformSchema,
+async function loadPlatformSchemas(): Promise<Array<PlatformSchemaShape>> {
```
silverbucket, line 79: this turned in to a very long function, this needs to be much better commented as to what it does and whats going on at various stages of the code flow.
and what it returns... jsdoc?

packages/schemas/src/context.ts
silverbucket, line 1: needs to be well documented jsdoc

biome.json
```diff
@@ -13,7 +13,8 @@
       "!**/*.test.js",
       "!**/packages/examples/build",
       "!**/packages/server/res",
-      "!**/packages/schemas/src/schemas/json"
+      "!**/packages/schemas/src/schemas/json",
+      "!**/packages/schemas/src/generated"
```
silverbucket: this can be removed

stress-tests/artillery/processors/sockethub-client.js
```diff
@@ -7,6 +7,18 @@
 const SockethubClient = require("@sockethub/client").default;
 const { io } = require("socket.io-client");
 
+const AS2_CONTEXT = "https://www.w3.org/ns/activitystreams";
+const SOCKETHUB_CONTEXT = "https://sockethub.org/ns/context/v1.jsonld";
```
silverbucket: should these should be loaded via sockethub dynamically during connect? I don't like the client having to have a static list of contexts for each platform.

stress-tests/artillery/processors/sockethub-client.js
```diff
@@ -261,7 +276,8 @@ function sendFeedMessage(context, events, done) {
 
     const message = {
         type: "fetch",
-        context: "feeds",
+        "@context": contextFor("feeds"),
+        platform: "feeds",
```
silverbucket: I don't like the redundancy of needing to write `contextFor("a platform name")` then also `platform: "a platform name"`...

Although I can understand the benefit of that platform property existing for readability, clarity, logging, ease of routing and ease of understanding for humans looking at the activitystream

What if, if the platform prop is not defined, it's automatically filled in. in the same way the activitiy streams library might expand an object that was previously a string, to ensure it;s always an object on looking, it will also ensure the platform prop always exists and is always a string, it will get the name through the proper channels.

packages/server/src/platform-instance.ts
```diff
@@ -274,6 +274,16 @@ export default class PlatformInstance {
         }
     }
 
+    /**
+     * Remove internal-only transport metadata before returning payloads to clients.
+     */
+    private toExternalPayload(payload: ActivityStream): ActivityStream {
+        const external = payload as InternalActivityStream;
+        delete external.sessionSecret;
+        delete external.platform;
```
silverbucket: I never suggested the platform property be deleted. I said it should not be required for a user to enter when composing an activity stream, and if it's not, it's inferred and populated to guarantee that any other place along the route (including any responses back to the client) can expect the platform property to exist and be a string. in the same way the activity streams package ensures, even if an actor is a string, that it expands it to an object so any other point along the route we can always safely access it.

packages/server/src/sockethub.ts
```diff
@@ -118,12 +146,23 @@ class Sockethub {
...
         sessionLog.debug("socket.io connection");
+        const platformRegistryPayload = this.buildPlatformRegistryPayload();
 
         // Rate limiting middleware - runs on every incoming event
         socket.use((event, next) => {
             this.rateLimiter(socket, event[0], next);
         });
 
+        // Send platform registry metadata to clients immediately and on-demand.
+        socket.emit("platforms", platformRegistryPayload);
```
silverbucket: all scheme information the client needs

packages/server/src/sockethub.ts
silverbucket, line 1: better jsdoc

packages/server/src/bootstrap/init.ts
```diff
@@ -35,45 +36,51 @@ function getExecutablePath(): string {
 export function printSettingsInfo(
     version: string,
     platforms: Map<string, PlatformStruct>,
+    options?: {
+        log?: (message?: unknown, ...optionalParams: Array<unknown>) => void;
```
silverbucket, line 40: explain why this is needed now and we have a logInfo object

packages/client/src/sockethub-client.ts
```diff
@@ -28,6 +34,35 @@ interface CustomEmitter extends EventEmitter {
...
+        sockethub?: string;
+    };
+    platforms?: Array<PlatformRegistryEntry>;
+}
+
+const AS2_CONTEXT_URL = "https://www.w3.org/ns/activitystreams";
+const SOCKETHUB_CONTEXT_URL = "https://sockethub.org/ns/context/v1.jsonld";
+const SOCKETHUB_PLATFORM_CONTEXT_BASE_URL =
+    "https://sockethub.org/ns/context/platform";
+const SOCKETHUB_PLATFORM_CONTEXT_VERSION = "v1";
```
silverbucket: still knows too much and is making too many assumptions.
all context data should be received from the sockethub server on connect. "schemas" event?

packages/server/src/sockethub.ts
silverbucket: Addressed in 5534f1bd. The server now emits all schema/context data the client needs on connect via the `schemas` event: `contexts.as`, `contexts.sockethub`, and per-platform schema/context metadata. The client now consumes this payload instead of constructing URLs locally.

packages/client/src/sockethub-client.ts
silverbucket: Addressed in c08c21fe (plus 5534f1bd). The client no longer hardcodes AS/Sockethub/platform context URLs. It now gets base contexts + platform metadata from the server `schemas` event, and `contextFor(platform)` is server-registry driven only (throws before schemas are loaded).

packages/server/src/sockethub.ts
silverbucket: Addressed in 5534f1bd and earlier docs pass. Added JSDoc around the new schema-registry flow in `packages/server/src/sockethub.ts` (payload builder + connection handler responsibilities) so the routing/metadata behavior is easier to follow.

packages/server/src/bootstrap/init.ts
silverbucket: Addressed in c08c21fe. Added an inline comment in `packages/server/src/bootstrap/init.ts` explaining why `logInfo`/`options.log` exists: testability (inject logger/exit shims so `printSettingsInfo` can be asserted without writing to stdout or exiting).

packages/client/src/sockethub-client.ts
```diff
@@ -166,14 +310,44 @@ export default class SockethubClient {
...
+                    // before they are emitted over the socket.
+                    const validationError = this.validateActivity(
+                        outgoing as ActivityStream,
+                    );
+                    if (validationError) {
+                        throw new Error(
+                            `SockethubClient validation failed: ${validationError}`,
+                        );
+                    }
+                }
```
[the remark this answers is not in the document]
silverbucket: Fixed in eb36fe4b.

Updated `SockethubClient` emit wrapper so outbound validation failures now follow socket callback semantics when a callback is provided:
- if callback is present, it receives `{ error: "SockethubClient validation failed: ..." }` and emit returns without sending payload
- if no callback is provided, behavior remains a thrown Error

Also added regression coverage in `packages/client/src/sockethub-client.test.ts` (`returns validation error via callback when registry is loaded`) to assert:
- no sync throw with callback
- callback receives formatted validation error
- invalid payload is not emitted to underlying socket.

packages/logger/src/index.ts
```diff
@@ -16,6 +16,33 @@ let hasLoggedInit = false;
...
+                    return "[Circular]";
+                }
+                seen.add(innerValue);
+            }
+            return innerValue;
+        });
+    } catch {
+        return '"[Unserializable]"';
+    }
+}
```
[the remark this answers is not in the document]
silverbucket: Added in 4c330f80.

I added regression coverage for the Error serialization path in `packages/logger/src/index.test.ts` (`serializes Error metadata with name/message/stack in log output`).

The test asserts the formatted log message includes:
- `"name":"Error"`
- `"message":"boom"`
- `"stack"`

So this path is now explicitly covered.

packages/activity-streams/src/activity-streams.ts
```diff
@@ -87,14 +87,6 @@ const baseProps = {
...
 } as const;
-const rename: Record<string, string> = {
-    "@id": "id",
-    "@type": "type",
-    verb: "type",
-    displayName: "name",
-    objectType: "type",
-    platform: "context",
-};
 const expand = {
```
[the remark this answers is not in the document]
silverbucket: Confirmed and intentional for this PR.

This migration removes AS1/legacy alias rewriting (`verb`, `objectType`, `@type`, `@id`, etc.) and standardizes on canonical AS2 fields only.

On your 3 checks:
1. Internal code paths in this PR were migrated to canonical AS2 payloads.
2. Client guidance is now centered on canonical payload composition (including `contextFor(...)` + server-provided schema/context registry).
3. Validation/normalization failures are surfaced with explicit `SockethubClient validation failed: ...` messages, and legacy alias handling is no longer performed.

No compatibility path is retained here by design for v5.

PR #1058, opened 2026-04-17
feat(platform-xmpp): Add token-based authentication option

silverbucket:
## Summary

- Allows XMPP credentials to supply either `password` or `token` (exclusive via JSON Schema `oneOf`), keeping all existing password flows unchanged.
- When `token` is provided, it is placed in the SASL PLAIN password slot by `buildXmppCredentials`, enabling auth against ejabberd `mod_auth_token`, Prosody `mod_tokenauth`, and similar modules — no `@xmpp/client` upgrade needed.
- Adds unit tests for both credential paths and a new AJV-based schema test covering password-only, token-only, both-present, and neither-present cases.

## Test plan

- [x] `bun test` in `packages/platform-xmpp` — 62/62 pass
- [x] `bun run lint` clean at repo root
- [x] `bun run doc` regenerates `API.md` with the new JSDoc token example
- [ ] Live end-to-end against an ejabberd server with `mod_auth_token` enabled (requires real deployment)


raucao:
There is no ejabberd `mod_auth_token`. I think we should go a bit slower with merging LLM code without human reviews.

General question: if the token is sent as the password anyway, what is the benefit of adding extra code to Sockethub, when a client supporting OAuth can just send the token as a password itself?

PR #1088, opened 2026-05-31
refactor(activity-streams): remove object registry and move helpers to schemas

silverbucket:
## Summary

This PR removes the `@sockethub/activity-streams` package and the server-side `activity-object` shorthand protocol. Activity stream normalization and development-time property linting now live in `@sockethub/schemas`, driven by platform message schemas from the server registry.

Net result: **−1,154 lines** across 66 files (952 additions, 2,106 deletions). The architecture is simpler — no process-wide object store, no separate ActivityStreams factory package — while platform-specific object properties (e.g. `xmpp:replace`) are still discovered from schemas and enforced at runtime.

## How it started

The branch began as **dynamic extension registration**: platform message JSON Schemas would declare which object properties are valid per type, and the client would register those after receiving the schema registry on `ready()`.

Early commits kept this logic inside `@sockethub/activity-streams`, extending `ASFactory` with schema-derived `registerObjectProps()` and tightening instance isolation so per-manager state did not leak across clients in the same process.

## Where it pivoted

The intermediate approach still carried complexity from the old model:

- A process-visible object registry (`activity-object` socket events, expand/create middleware)
- Duplication between `@sockethub/activity-streams` and `@sockethub/schemas`
- Risk of module-global state in a package used by both client and server

Commit `214158f6` is the main pivot: **delete `@sockethub/activity-streams` entirely** and move the useful parts into `@sockethub/schemas`.

**Removed:**
- `@sockethub/activity-streams` package
- `expand-activity-stream` and `create-activity-object` server middleware
- `activity-object` socket handler
- `packageConfig` in server config (never wired at runtime; tracked in #1097)

**Replaced with:**
- `createActivityStreamProcessor()`, `normalizeActivityStream()`, and `extractObjectPropertyExtensionsFromMessageSchema()` in `@sockethub/schemas`
- `normalize-activity-stream` middleware (string actor/target/object refs → minimal objects)
- Clients send full actor objects directly

### Client (`@sockethub/client`)

- Private `streamProcessor` per `SockethubClient` instance
- On `schemas` event: registers platform object-property extensions from each platform's message schema, then becomes `ready`
- Inbound/outbound `message` and `credentials` pass through the processor for normalization and unknown-property linting
- Reconnect replay unchanged in intent: `credentials`, `connect`, and `join` only

### Server (`@sockethub/server`)

- `normalize-activity-stream` middleware runs before AJV validation
- Validation order preserved: normalize → validate (base + platform-specific)

### Schemas (`@sockethub/schemas`)

- New helpers exported for client, server, and tests
- AJV error messages improved for `additionalProperties` violations (includes the offending key)
- Browser-compatible build split for Vite/examples consumption

### Examples and integration

- Examples updated to send complete `actor` objects (`id`, `type`, `name`) on every outbound event
- Integration tests no longer depend on `ActivityStreams.Object.create`
- Browser integration CI hardened with docker-compose image-pull retries (Ergo timeout fix)

### Other fixes on this branch

- Feeds platform: defensive parsing for entries missing `link`/`meta`; empty-string fallbacks for `title`/`content`
- Examples: submit edited JSON from credential text areas; `JSON.parse` error handling; credentials ack wired to returned promise

## Breaking changes

- **`activity-object` protocol removed** — no server-side object creation/expansion via socket events
- **Actors must be full objects** — string actor refs are normalized server-side, but clients should send `{ id, type, name? }` explicitly
- **`@sockethub/activity-streams` removed** — depend on `@sockethub/schemas` instead
- **`packageConfig` removed** from server defaults and config schema

## Summary by CodeRabbit

* **New Features**
  * Platform-specific object properties are now accepted and passed through when allowed by server schemas.
  * Clients and examples now send full actor objects (id, type, optional name) and await client readiness before emitting events.

* **Bug Fixes**
  * Activity stream validation/normalization improved to accept registered extensions and namespace properties (e.g., xmpp:replace) without failing.

* **Documentation**
  * Guides updated to reflect schema-driven property handling and credentials emit acknowledgement pattern.


silverbucket (commented):
All review comments addressed in the latest commit (68ba6b40):

**Major: Module-global state in ASFactory** — Resolved. `customProps`, `failOnUnknownObjectProperties`, `warnOnUnknownObjectProperties`, and `specialObjs` are now declared inside `ASFactory()`. `validateObject`, `registerObjectProps`, `matchesCustomProp`, `getAllowedObjectProps`, `Stream`, and the `Object` manager are all closures over per-instance state, so registrations from one manager can no longer leak into another. The `objs` map is also instance-scoped. A side-effect was that the `warn test` block's ``rejects nondefined special types`` test was only passing due to the module-global mutation bug; that test now explicitly sets `failOnUnknownObjectProperties: true`.

**registerActivityStreamsSchemaProps outside try/catch** — Moved inside the existing resilient block so a malformed messages schema cannot abort registry application for remaining platforms.

**Use `test` instead of `it`** — Changed the newly added test cases in `sockethub-client.test.ts`, `index.test.ts`, and `schema.test.ts` to use `test()`. The rest of the file uses `it` as a global (pre-existing pattern); only the PR-introduced tests are changed per the minimal-fix guidance.

**Local ActivityStreamsExtensions interface** — This was removed in the `test(client): avoid real platform schema fixture` commit prior to review. Not present in the current code; the local `JsonSchemaLike` interface serves a different purpose (schema traversal) and has no equivalent in `@sockethub/schemas`.

**Double normalization / shallow copy of extensions** — These comments applied to the `4efba6c9` version of the code. The `extensions` field was removed from `PlatformRegistryEntry` in the `1505d5a3` commit, so these are no longer present.

**Merge conflict** — Resolved in `schema.test.ts`, keeping both the branch-added `test("is a valid Sockethub platform schema")` and the master-updated `test("accepts password-only credentials")`.

PR #1104, opened 2026-06-07
ci: run integration tests nightly on master

silverbucket:
## Summary

Adds a nightly schedule (3:00 UTC) and `workflow_dispatch` to the Integration workflow, running the existing jobs against master. This catches environment drift — unpinned docker images (`redis`, `prosody:latest`, `ergo:stable`), `bun-version: latest`, runner updates — that no PR trigger would surface. Also scopes workflow permissions to `contents: read`.

PR-triggered runs are unchanged.

## Why the original tag matrix was dropped

The earlier version of this PR ran the nightly against three refs (master, latest-stable, latest-tag) by checking out old tags under master's workflow steps. That was structurally broken:

- `v4.1.0` (latest-stable) predates the bun/`integration:*` script layout entirely — its matrix leg could never pass, and tags are immutable, so the failure is permanently non-actionable
- tag legs break silently whenever master's workflow drifts from a tag's repo layout, which happened twice during this PR's lifetime (#1106, #1108)
- master already removed this same setup-job + resolver-script pattern from `test-npm-package.yml` in #1106

Verifying that *released artifacts* still work is a separate concern that belongs in the Test NPM Package workflow (which takes a `ref` input), against the latest release only.


silverbucket:
The review above targets the pre-slim revision (range ended at f34fc125). Since then, 765c651e reduced this PR to just the schedule/dispatch triggers and `permissions: contents: read`:

- `scripts/lib/git-release-tags.mjs` and `scripts/resolve-nightly-integration-refs.mjs` (both sorting and error-handling comments) were deleted — the tag matrix is gone.
- The setup-bun SHA-pinning and composite-action suggestions describe master's pre-existing job steps, which this PR no longer modifies. Both are reasonable repo-wide cleanups but belong in a separate chore PR rather than this 8-line trigger change.

@coderabbitai review

PR #1107, opened 2026-06-08
fix(platform-xmpp): use unique IDs for attendance queries

### LINKED ISSUE #1090: fix(platform-xmpp): use unique IDs for attendance queries
## Problem

The `query` method in `packages/platform-xmpp/src/index.js` uses a hardcoded static ID `"muc_id"` for attendance (disco#items) IQ queries:

```js
this.__xml("iq", { id: "muc_id", type: "get", ... }, ...)
```

If two attendance queries are in-flight simultaneously, both responses arrive with `id="muc_id"` and both are routed to `notifyRoomAttendance` — which is likely correct in that case, but the lack of unique IDs means there is no way to correlate a response to a specific request, and the routing relies on a collision-prone constant.

By contrast, room-info queries (added in #916) use `room_info_${randomUUID()}`, which is the correct approach.

## Suggested Fix

Replace `id: "muc_id"` with a unique ID using a `attendance_${randomUUID()}` pattern, and update the routing check in `incoming-handlers.js` (`stanza.attrs.id === "muc_id"`) to use the same prefix-based approach as room-info queries.

## Context

Noticed during review of #916, which introduced room-info queries using UUID-based IDs in the same function. This inconsistency was pre-existing.

mvanhorn:
## Summary

XMPP attendance IQ stanzas now use unique per-query IDs instead of a single fixed `muc_id`, so concurrent attendance/room queries no longer collide on their response correlation.

## Why this matters

The XMPP platform sent attendance/room-info IQ queries with a hardcoded `id` (#1090). Multiple in-flight queries shared one ID, so responses could be correlated to the wrong request. This generates a unique `attendance_`-prefixed ID per query and matches incoming responses against it, mirroring the pattern the room-info path already uses.

## Testing

- `bun test packages/platform-xmpp/` passes (99 tests); the stale fixed-ID assertions were updated to assert the dynamic `attendance_` shape.

Closes #1090

## Summary by CodeRabbit

* **Chores**
  * Improved attendance query request/response matching by using unique identifiers instead of static IDs, ensuring more reliable routing of attendance-related responses.
  * Updated test coverage to validate new attendance query ID generation and response handling logic.

### DIFF
--- a/packages/platform-xmpp/src/incoming-handlers.test.ts
+++ b/packages/platform-xmpp/src/incoming-handlers.test.ts
@@ -2,10 +2,12 @@ import { beforeEach, describe, expect, test } from "bun:test";
 import sinon from "sinon";
 
 import * as schemas from "@sockethub/schemas";
+import type { XmppClientInstance, XmppElement } from "@xmpp/client";
 import parse from "@xmpp/xml/lib/parse.js";
 
 import { IncomingHandlers } from "./incoming-handlers.js";
 import { stanzas } from "./incoming-handlers.test.data.js";
+import XMPP from "./index.js";
 import { PlatformSchema } from "./schema.js";
 import type { XmppHandlerSession } from "./types.js";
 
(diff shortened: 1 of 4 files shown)


mvanhorn:
Appreciate you merging this, @silverbucket. Unique IDs for attendance queries close out that XMPP response mixup.

silverbucket:
Thanks for your contribution @mvanhorn, much appreciated!

PR #1108, opened 2026-06-08
ci: de-flake browser integration suite (metadata fixture, multiclient delivery barrier, contract job split)

silverbucket:
## Summary

De-flakes the integration **browser** CI job, which began failing intermittently once #1098 added the contract tests to it. Several independent flaky failure modes were in play; this PR addresses each and isolates them so one can't block the others.

### 1. Metadata contract test — remove the external network dependency
- Add a local `metadata-test.html` fixture served by the examples app.
- Point the metadata contract test at the local fixture instead of `https://sockethub.org`, and assert deterministic title/url/description values.
- Use a `127.0.0.1` fixture URL when `config.sockethub.url` is localhost (IP literals satisfy `open-graph-scraper`'s URL validator without touching production scraper options).

### 2. Multiclient XMPP test — fix the cross-client delivery race
- A MUC only delivers a message to occupants already present when it is broadcast (no replay to late joiners). The test sent a single message immediately after join, so a slightly-late occupant could permanently miss it.
- Gate the broadcast on a **resend-until-delivered barrier**: generalize the existing `waitForStableRoomDelivery` helper to take a sender, and have the cross-client test delegate to it. It re-sends a freshly tagged message (with backoff) until every peer (excluding the sender) receives one, converging on confirmed room-wide delivery and failing only if delivery never completes.
- (An attendance-query barrier via the platform's `disco#items` feature was attempted first, but prosody returns an empty occupant list for the room in this setup, so it can't confirm membership — see notes.)

### 3. Isolate contract tests in their own CI job
- Move the platform contract tests (feeds/dummy/metadata) out of the `browser` job into a dedicated `contract` job that starts only Sockethub + Redis (no XMPP/IRC).
- A flaky XMPP/IRC delivery race in the `browser` job no longer blocks unrelated contract changes, and vice versa.

### 4. Browser job startup flake
- Headless Chrome intermittently failed to start within web-test-runner's 30s default ("unable to create and start a test page") while sharing the runner with the full docker stack. Raised `browserStartTimeout`/`testsStartTimeout`.

### Other
- Multiclient mismatched-credentials test now waits for `SockethubClient.ready()` before emitting credentials, without increasing the credential ACK timeout.
- `persist-credentials: false` on all `integration.yml` checkout steps (none of these jobs need the token).

## Notes
- **SSRF (deferred to #1109):** `platform-metadata` passes the client-controlled `actor.id` to `open-graph-scraper` with no private/loopback filtering. This is a pre-existing gap and a proper fix conflicts with the local-fixture approach here, so it's split into a focused security PR (#1109).
- **Attendance feature gap:** the room-attendance (`disco#items`) query returns an empty occupant list against the integration prosody, so it currently has no usable integration coverage; worth a separate look.

## Validation
- `bun run lint`
- `bun run build`
- Browser/contract behavior verified via CI (local Docker is unavailable in this environment).

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -107,9 +107,6 @@ jobs:
       - name: Basic Browser Integration Tests
         run: XMPP_HOST=prosody bun run integration:browser:basic
 
-      - name: Platform Response Contract Tests
-        run: XMPP_HOST=prosody bun run integration:browser:contract
-
       - name: Sockethub Logs - Basic
         if: always()
         continue-on-error: true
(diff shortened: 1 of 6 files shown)


silverbucket:
Thanks for the reviews — here's how each item was addressed:

**Metadata SSRF (Cursor, Medium)** — valid, but it's a *pre-existing* gap in `platform-metadata` and not introduced by this PR. A proper fix (blocking private/loopback/link-local targets) directly conflicts with this PR's approach of serving a local `127.0.0.1` fixture to remove the external-network flake. Rather than weaken the fix with a test-only bypass or undo the local fixture, I've split it into a focused security PR: #1109. This PR stays scoped to de-flaking CI.

**URL validator / `require_tld` (CodeRabbit, Major)** — already resolved: the fixture uses `127.0.0.1`, and `open-graph-scraper`'s validator treats IP literals as valid (`require_tld` doesn't apply), so no `urlValidatorSettings` change is needed.

**`persist-credentials: false` (CodeRabbit, Major)** — done, applied to all checkout steps in `integration.yml` (not just the new `contract` job) for consistency. None of these jobs need the token.

**Pin actions to SHA (CodeRabbit, Critical)** — the `contract` job intentionally matches the `@v3`/`@v2.0.1` tags used by every other job in this same workflow; pinning only the new job would be inconsistent. SHA-pinning is a worthwhile file-wide (or repo-wide) modernization, but it's out of scope for a CI-de-flaking PR and better done in one consistent pass.

**`sleep 5` → readiness check** — the `contract` job already follows the existing `browser` job pattern: `sleep 5` followed by `nc -zv` readiness probes on each required port.

Also fixed the actual browser-job failure, which was unrelated to the test logic: headless Chrome intermittently failed to start within web-test-runner's 30s default ("unable to create and start a test page") while sharing the runner with the full docker stack. Raised `browserStartTimeout`/`testsStartTimeout`.

silverbucket:
Filed the room-attendance gap noted above as #1110. Evidence from this PR's CI run: the attendance query round-trips and the server logs `got room attendance list`, but `object.members` comes back `[]` even though the 5 clients are provably in the room (they exchange messages fine) — so the feature has no working integration coverage and likely doesn't enumerate occupants for the common semi-anonymous/non-moderator case. That's why this PR uses a resend-based delivery barrier instead.

Issue #1109, opened 2026-06-08
Harden platform-metadata against SSRF (block private/loopback/link-local targets)


silverbucket:
## Problem

`packages/platform-metadata/src/index.ts` passes the client-controlled `job.actor.id` directly to `open-graph-scraper` with no filtering:

```ts
fetch(job, cb) {
    ogs({ url: job.actor.id })  // actor.id is client-controlled, no allowlist / private-IP check
    ...
}
```

Because `actor.id` comes from the client and is not validated against private/loopback/link-local ranges, any client that can send a metadata `fetch` can make the Sockethub server issue requests to internal services (SSRF). Flagged by Cursor review on #1108.

## Proposed fix

- Reject private (RFC 1918), loopback (127.0.0.0/8, ::1), link-local (169.254.0.0/16, fe80::/10), and other non-public targets before fetching, resolving the hostname first to avoid DNS-rebinding.
- Consider a configurable allowlist for trusted internal targets (and to keep integration tests able to fetch a local fixture).
- Restrict protocols to http/https (open-graph-scraper default).

## Context / scope note

Surfaced while reviewing #1108, which de-flakes the metadata integration test by serving a local `127.0.0.1` fixture. The fix conflicts with that loopback fixture approach, so it was intentionally deferred from #1108 (a CI-de-flaking PR) to this focused security PR. See https://github.com/sockethub/sockethub/pull/1108 for discussion.

silverbucket:
Closing as done — implemented in #1138 and #1140: platform-metadata now fetches through a guarded dispatcher (`packages/util/src/net/dispatcher.ts`) that blocks private/loopback/link-local/cloud-metadata addresses at the DNS-lookup/connect layer (closing the DNS-rebinding gap), enforces the guard across redirect hops, caps body size, and restricts protocols to http/https. The test escape hatch exists as `allowPrivateAddresses` in packageConfig, and platform-feeds received the same treatment in #1133/#1139.

PR #1111, opened 2026-06-08
fix(platform-metadata): accept TLD-less hosts so localhost fixtures are fetchable

silverbucket:
## Summary

Fixes the `browser-contract` failure in the **Test NPM Package** workflow (run [27139752074](https://github.com/sockethub/sockethub/actions/runs/27139752074)) that surfaced after #1108 merged. CI was green but the packaged run failed:

```
❌ metadata platform > returns page object fields...
   Error: metadata fetch: connect ECONNREFUSED 127.0.0.1:10550
```

## Root cause

#1108 rewrote the metadata fixture URL `localhost` → `127.0.0.1` purely to satisfy `open-graph-scraper`'s URL validator (`require_tld: true` rejects bare `localhost`). But Sockethub listens on `config sockethub:host` (default `localhost`), which on the GitHub runner resolves to **IPv6 `::1`** — so the metadata platform's IPv4 `127.0.0.1` fetch was refused. In docker CI, Sockethub binds to all interfaces, so `127.0.0.1` happened to work — which is why the PR's own CI was green. The feeds contract test (same run) used `localhost` and passed, confirming `localhost` is the universally reachable address.

## Fix

- `platform-metadata`: relax the scraper's URL validator to allow TLD-less hosts (`require_tld: false`). Since ogs *replaces* rather than merges `urlValidatorSettings`, the other defaults are restated. This is the approach CodeRabbit originally recommended (option a) on #1108.
- Contract test: fetch the fixture via `config.sockethub.url` (`localhost`) — reachable in both docker and host environments — instead of the IPv4-only `127.0.0.1` rewrite.
- Fixture: `og:url` updated to the `localhost` form to match.

## SSRF note (#1109)

`require_tld: false` lets the metadata platform fetch bare hostnames (e.g. `localhost`, intranet names). The default validator **already** accepts bare IPs, so IP-based SSRF was always possible and this does not introduce a new class of exposure on its own. The proper mitigation remains #1109 (block private/loopback targets by **resolved IP**, which must now also cover hostname-resolved IPs).

## Validation
- `bun run lint`, `bun run build` (platform-metadata)
- Confirmed against the installed `open-graph-scraper@6.11.0` that `isUrlValid("http://localhost:10550/metadata-test.html", { ...defaults, require_tld: false })` returns `true` (and `false` with the defaults — the original failure mode).
- Full packaged verification runs in the Test NPM Package workflow.

## Bug Fixes

* Fixed metadata URL validation and configuration handling to properly support localhost development environments, enabling consistent behavior across local development setups.

### DIFF
--- a/integration/contract/platform-responses.integration.js
+++ b/integration/contract/platform-responses.integration.js
@@ -135,10 +135,12 @@ describe(`Platform response contracts at ${config.sockethub.url}`, () => {
 
     describe("metadata platform", () => {
         it("returns page object fields the examples metadata view expects", async () => {
-            const fixtureUrl = `${config.sockethub.url.replace(
-                "://localhost:",
-                "://127.0.0.1:",
-            )}/metadata-test.html`;
+            // Fetch the fixture via the same base URL the rest of the suite uses.
+            // An earlier revision rewrote localhost -> 127.0.0.1 to satisfy the
+            // scraper's URL validator, but that broke when Sockethub binds only to
+            // IPv6 localhost (::1) — the IPv4 fetch was refused. The metadata
+            // platform now accepts TLD-less hosts, so localhost works everywhere.
+            const fixtureUrl = `${config.sockethub.url}/metadata-test.html`;
             const msg = await emitWithAck(
                 sc.socket,
                 "message",
--- a/packages/examples/static/metadata-test.html
+++ b/packages/examples/static/metadata-test.html
@@ -7,7 +7,7 @@
     <meta property="og:title" content="Sockethub Metadata Test">
     <meta property="og:description" content="Local metadata fixture for integration tests">
     <meta property="og:type" content="website">
-    <meta property="og:url" content="http://127.0.0.1:10550/metadata-test.html">
+    <meta property="og:url" content="http://localhost:10550/metadata-test.html">
   </head>
   <body>
     <h1>Sockethub Metadata Test</h1>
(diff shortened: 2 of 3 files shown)


silverbucket:
## Verified against the Test NPM Package workflow

Dispatched this workflow against the branch (the only CI that reproduces the original `ECONNREFUSED` failure — normal PR CI runs in docker, which binds all interfaces and hides it).

**Run [27141407911](https://github.com/sockethub/sockethub/actions/runs/27141407911) (suite=all, both runtimes):**

| suite | node | bun |
|---|---|---|
| browser-contract (the fix) | ✓ | ✓ |
| browser-multiclient | ✓ | ✗ (unrelated — see below) |
| all others incl. IRC | ✓ | ✓ |

- **node: fully green.**
- **bun:** the only failure was `Performance and Load Testing > rejects connection with mismatched credentials` — `Timeout waiting for ack after 2000ms for event=credentials`. That is a **pre-existing flaky test** (it passed on the original master run [27139752074](https://github.com/sockethub/sockethub/actions/runs/27139752074) and passed under node here), unrelated to this metadata change. It's the credential-ACK race #1108 already tried to harden.

The reported regression — `browser-contract` failing with `connect ECONNREFUSED 127.0.0.1:10550` — is **fixed** and confirmed on both runtimes.

### Two unrelated issues surfaced during verification
1. **Harness bug:** `scripts/test-installed-version/services.js` sets `needsIrc = suite === "all"`, but the `browser` suite runs the 4 IRC browser tests — so `suite=browser` runs them without starting ergo and they fail-fast. One-line fix: include `browser` in `needsIrc`.
2. **Flaky test:** the mismatched-credentials test still races on the 2s credential ACK under bun.

Happy to fold #1 (and/or tackle #2) into this PR or file them separately.

Issue #1120, opened 2026-06-10
Define per-platform outbound object/message schemas and validate messages on send


silverbucket:
## Motivation

Sockethub validates **inbound** messages (client → server) against each platform's message schema, but **outbound** messages (platform → client) are sent with no schema validation. Platform responses can therefore cross the server/client boundary malformed — a typo or missing field in an outgoing object goes undetected, and there's no schema a platform developer must satisfy when adding a new outbound shape.

## Why a naive fix doesn't work (findings from #1119)

We explored validating outbound messages in `PlatformInstance.sendToClient` and found the current schemas are **not outbound-complete**, so any strict validation drops legitimate responses:

- The base ActivityStreams envelope validates `object` against registered object types, but outbound object types aren't registered — e.g. metadata returns `object: { type: "page" }` and `page` is **not** in `ObjectTypesSchema` → the metadata response would be dropped.
- Per-platform *message* schemas only describe **inbound commands** (e.g. feeds `type` enum is `["fetch"]`), while responses use other types (`collection`, `post`, `page`, presence `message`) → validating responses against the inbound schema rejects them.
- Actor sub-schemas deep-validate (e.g. `feed`/`website` actors require IRI `id`), and the server's error path even injects an invalid actor type (`{ type: "unknown" }`, not in `ActorTypesList`).

So outbound validation and outbound schemas are one cohesive unit — you can't safely enforce validation until the schemas exist.

## Proposed shape

1. **Per-platform outbound schemas.** Alongside each platform's inbound `messages`/`credentials` schema, define an `outbound` (response) schema describing the message/object types that platform can emit (e.g. feeds → `collection`/`post`, metadata → `page`, xmpp → presence/`message`). New features require the developer to add/extend the outbound schema.
2. **Register the outbound object types** (`page`, `post`, etc.) in `ObjectTypesSchema`, or scope object validation to the platform's outbound schema.
3. **Enforce on send.** In `sendToClient`, after `toExternalPayload`, validate the message against the platform's outbound schema; drop + log with context (session, platform, type) on failure. Fix the error-path actor type (`unknown` → a registered type such as `service`).
4. **Tests.** Each platform gets outbound-schema fixtures; the server gets a test that malformed outbound messages are dropped and well-formed ones pass.

## Benefits
- Catches malformed/typo'd outgoing objects with clear errors (the gap that's currently unchecked).
- Forces a documented contract for every outbound shape, improving platform test coverage.
[...]

PR #1122, opened 2026-06-10
feat: strict per-platform outbound responses schemas + enforce on send (#1120)

silverbucket:
Implements the core of #1120: per-platform **outbound** (`responses`) schemas, fully strict (`additionalProperties: false`), enforced in `PlatformInstance.sendToClient`. Builds on the warn-only infra from #1121.

## What this does
Each protocol platform now declares the exact shape of every message it emits, and the server **drops** (logs at error) any outbound message that doesn't match — symmetric with inbound validation.

- **platform-feeds** — `collection` message + `post` items (feed-channel actor + article/note object).
- **platform-metadata** — `page` response (website actor + open-graph-derived page object incl. the ogImage array shape).
- **platform-xmpp** — every outbound message type (connect/close/error/update/send/query/request-friend/join/leave) and object type (message/presence/attendance/room-info/connect), incl. the top-level `error` field and dynamic room-info form maps.
- **platform-irc** — every outbound type (update/send/join/leave/add/remove) and object type (message/me/presence/topic/address/relationship).
- **server** — `sendToClient` flips warn→**enforce** (drop+log on mismatch). Fixes the injected/error-path actor type `unknown`→`service` (`unknown` isn't a valid actor type).

## Validation — fully strict, verified against real output
Per the agreed approach, schemas are `additionalProperties: false` (no silent passthrough), and every schema is unit-tested against the platform's **real** output (not hand-built fixtures):
- feeds: real `buildFeedItem` output from the RSS fixtures.
- metadata: real `fetch()` output (open-graph-scraper mocked).
- xmpp: every emitted message in the `incoming-handlers` fixtures.
- irc: the canonical `@sockethub/irc2as` translation outputs.

All unit suites green locally: schemas 62, feeds 23, metadata 4, xmpp 118, irc 79, server 121; lint + full build clean. The integration/contract/browser suites (CI) exercise real end-to-end flows and are the backstop for any shape not covered by fixtures.

## Scope notes
- **`error` envelopes are exempt** from responses validation — they're a generic cross-cutting shape (type `error` + `error` string + injected actor), not a platform protocol response; validating them against per-platform schemas would drop them.
- **platform-dummy is intentionally skipped** — it's an echo test-harness that deliberately passes arbitrary/malformed payloads, so a strict outbound schema would defeat its purpose. It remains a no-op (no `responses` schema).
- The base envelope / global object `oneOf` in `objects.ts` is **untouched** here. Outbound validation runs only the per-platform `responses` schema, so it never touches the base envelope. Splitting the `objects.ts` stop-gap is an inbound-only concern, tracked separately.

## Summary by CodeRabbit

* **Improvements**
  * Outbound message validation is now strictly enforced across platforms; invalid messages are dropped and logged instead of being emitted with warnings.
  * Error envelopes and certain failure notifications bypass schema validation so error reporting remains deliverable.
  * Error responses and error reporting now set the responding service as the actor when an actor is missing.

* **Tests**
  * Added validation tests covering canonical responses and negative cases for unknown types/fields.

### DIFF
--- a/packages/platform-feeds/src/index.test.ts
+++ b/packages/platform-feeds/src/index.test.ts
@@ -2,9 +2,42 @@ import { readFileSync } from "node:fs";
 import path from "node:path";
 import { beforeEach, describe, expect, test } from "bun:test";
 import type { ASCollection, PlatformSession } from "@sockethub/schemas";
+import {
+    addPlatformContext,
+    addPlatformSchema,
+    validateActivityStreamResponse,
+} from "@sockethub/schemas";
 import getPodcastFromFeed from "podparse";
+import feedsSchema from "./schema";
 import { RSSFeed } from "./index.test.data";
 import Feeds, { buildFeedItem, datesEqual } from "./index";
+
+addPlatformSchema(feedsSchema.responses, "feeds/responses");
+addPlatformContext("feeds", feedsSchema.contextUrl);
+
+// Wrap feed-item objects exactly as the feeds platform emits them: a single
+// `collection` message whose `items` are `post` entries. Used to validate real
+// parsed output against the strict outbound `responses` schema.
+function buildCollection(objects: Array<ReturnType<typeof buildFeedItem>>) {
+    return {
+        "@context": [feedsSchema.contextUrl],
+        type: "collection",
+        summary: "Test Feed",
+        totalItems: objects.length,
+        items: objects.map((object) => ({
+            "@context": [feedsSchema.contextUrl],
+            type: "post",
+            actor: {
+                id: "https://example.com/feed",
+                type: "feed",
+                name: "Example",
+                link: "https://example.com",
+                categories: [],
+            },
+            object,
+        })),
+    };
+}
 
 const FIXTURES_DIR = path.join(import.meta.dirname, "../test/fixtures");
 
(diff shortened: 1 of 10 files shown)


silverbucket:
Addressed the outside-diff review comment:

**Global platforms produced error actors with `id: undefined`.** Fixed in both error-actor paths — `reportError` and the `sendToClient` injection now fall back to the platform name (`this.actor ?? this.name`) so every `error` envelope carries a valid, id-bearing `{ id, type: "service" }` actor. The injection also no longer requires `this.actor` to be truthy, so actorless errors from global platforms are populated too. Added a test covering the global-platform path.

silverbucket:
Addressed both items from the latest review:

**1. Failure notifications were being dropped (the real issue).** You're right — on job failure, `handleJobResult` echoes the original request (an inbound type like `fetch`/`send`) plus an `error` field, which under enforce was validated against the *outbound* `responses` schema and dropped.

Re your suggestion to validate against inbound **or** outbound (either passes): it works, but I went a different way because either-passes silently bypasses outbound strictness wherever inbound/outbound share a type or inbound is permissive — e.g. **metadata** uses `fetch` for both, and its inbound `object` schema accepts any object, so a successful `page` response would pass the inbound check regardless and never be validated against the strict `page` schema.

Instead: **any message carrying an `error` field is a failure/error notification, not a protocol response, so it's exempt** (extending the existing `type: "error"` exemption). This keeps the request-echo failure shape (clients keep correlation) and doesn't weaken success-path validation, since successful responses never carry `error`.

**2. Added the `reportError()` regression test** asserting the `{ id: this.name, type: "service" }` fallback actor for a global platform, plus a test for the failure-echo exemption.

Server suite 124 pass, lint + build clean.

silverbucket:
Both points in the latest review are already addressed in `474712cf` (the review was generated against `a13fac9b`, one commit earlier):

- **Nitpick — `reportError()` fallback-actor regression test:** added (`platform-instance.test.ts`, "reportError emits a valid service actor for a global platform") — calls `reportError()` directly and asserts `{ id: this.name, type: "service" }`.
- **Failed-job validation drop:** fixed — messages carrying an `error` field (failure echoes) are now exempt from `responses` validation, so failure notifications are no longer dropped.

A re-review of `474712cf` should clear both.

Issue #1123, opened 2026-06-10
Split platform-specific object types out of the global objects.ts (inbound object ownership)


silverbucket:
Follow-up to #1120 / #1122. The outbound (`responses`) validation landed in #1122 without touching the base envelope. This issue covers the remaining, separable **inbound** concern: the global object-type registry in `packages/schemas/src/helpers/objects.ts`.

## Background
`ObjectTypesSchema` in `@sockethub/schemas` is a **static, hardcoded** registry (never populated by platforms). The base ActivityStreams envelope embeds it as `object: { oneOf: <every type> }`, so **inbound** validation (`validateActivityStream`) checks message objects against this global set. It was a stop-gap; object-type ownership should move to the platforms that emit/accept them.

Note: outbound validation does **not** use this — `validateActivityStreamResponse` runs only the per-platform `responses` schema. So this is purely an inbound refactor.

## Why it's separate / non-trivial
- The global set doubles as the **actor/target vocabulary** (`person`, `room`, `service`, `feed`, `website`, `platform`, `address`) via `validActorRefs`/`validTargetRefs`, which must keep resolving.
- `relationship` references `validObjectRefs` (all object types) internally.
- Removing the base envelope's object `oneOf` **loosens inbound validation for every platform at once** unless each platform's `messages` schema is first tightened to own its inbound objects. High blast-radius; will touch existing inbound tests.

## Proposed approach
1. Keep the **actor/target vocabulary** in `@sockethub/schemas` core (it's shared and stable).
2. Move platform-specific **object** definitions (`message`, `presence`, `attendance`, `room-info`, `relationship`, `topic`, `address`, `me`, `feed`, `page`/`post`, …) to the owning platforms, exposing genuinely-shared shapes as reusable building blocks exported from `@sockethub/schemas` (composed per platform — the option-1 approach agreed in #1120). The per-platform `responses` schemas added in #1122 already define many of these inline and can be a reference.
3. Tighten each platform's `messages` (inbound) schema to fully validate its objects.
4. Remove the global object `oneOf` from the base envelope (it becomes redundant once platforms self-validate).
5. Delete the now-unused object-only entries from `objects.ts`.
6. Verify inbound validation strictness is preserved per platform (unit tests + the contract/browser suites).

## Done when
- `objects.ts` holds only the shared actor/target vocabulary (+ any truly core/internal types like `heartbeat`).
- Inbound object validation is owned by each platform's `messages` schema, with no loosening vs. today.
[...]

PR #1124, opened 2026-06-10
refactor(schemas): per-platform inbound object validation + remove dead activity-object path (#1123)

silverbucket:
Addresses #1123: move inbound object-type ownership to the platforms and stop the base envelope from enumerating a global object vocabulary for validation — symmetric with the per-platform `responses` schemas from #1122. Also removes a dead validation subsystem.

## 1. Remove the dead `activity-object` validation path
`validateActivityObject` / `ActivityObjectSchema` and the `activity-object` branch in the server validate middleware were **unreachable** — the middleware is only ever constructed with `credentials`/`message`, no socket event emits `activity-object`, and the client never sends one (a relic of the old standalone-object-store concept). Removed the validator, schema, AJV registration, barrel exports, generator-script entry, orphaned JSON artifact, dead middleware branch, and its test data. `handleValidation` loses its now-dead `isObject` path. The `ActivityObject` **type** is kept (it's the type of the AS `object` field).

## 2. Platforms own inbound object validation
- **xmpp** `messages` schema gains strict (`additionalProperties:false`) inbound object defs: message/presence/attendance/room-info.
- **irc**: message/topic/address/attendance.
- **feeds/metadata** already constrained their inbound objects; **dummy** stays permissive (echo test-harness).
- **base envelope** (`activity-stream.ts`): `object` drops its global `oneOf` (now just `type:object`); `actor`/`target` still validate against the shared actor/target vocabulary.
- **objects.ts**: removed the now-unused `validObjectRefs`/`validObjectDefs` machinery.

## Scope note: `ObjectTypesSchema` retained (deliberately)
`ObjectTypesSchema` is **not** emptied here because it's a **live** dependency of the normalizer — `normalizeActivityStream` (server `normalize-activity-stream` middleware) and `createActivityStreamProcessor` (`@sockethub/client`) iterate it to build their known-object-property vocabulary. Removing the content types would make the normalizer treat `content`/`presence`/etc. as unknown properties — a live regression. Fully retiring `ObjectTypesSchema` requires reworking the normalizer to source object properties from the per-platform schemas; tracked as a follow-up.

## Tests
- schemas stream fixtures now exercise **per-platform** object validation via an object-aware test schema (validates the new mechanism).
- xmpp `incoming-handlers` validates outbound asobjects against the **responses** schema (they're outbound — previously mis-validated against the inbound validator, which only passed because the global object schema was lenient).
- xmpp/irc gain direct **inbound** object-validation unit tests.
- All suites green locally: schemas 40, server 118, xmpp 121, irc 82, feeds 23, metadata 4, client 33; lint + full build clean. CI contract/browser cover inbound end-to-end.

## Summary by CodeRabbit

* **New Features**
  * Platforms now validate inbound message objects against platform-specific object types (message, topic/presence, attendance, room-info).

* **Refactor**
  * Object validation moved from a generic activity-object surface into platform-specific message schemas.
  * Validation middleware simplified to operate on ActivityStream messages only.

* **Tests**
  * Expanded tests for inbound message validation and rejecting unknown fields.

* **Chores**
  * Removed activity-object exports/validation surface and stopped exporting the activity-object schema.


silverbucket:
Assessed the outside-diff comments:

**1. "Critical: race condition — async init not awaited" — valid concern, but not actually critical; fixed properly.**
The fire-and-forget `getInitObject().then(initObj = …)` is **pre-existing** (since the Bun migration #886), and in practice it does *not* fail: production `await`s `getInitObject()` before `listener.start()`/`io.on("connection")` (sockethub.ts), so init is fully resolved/cached before any socket connects, and the `.then` microtask completes before any client message arrives over the network. So CodeRabbit's "will always fail unless passedInitObj is provided" isn't accurate.

CodeRabbit's proposed fixes don't fit: making `validate()` `async` breaks `.use(validate(...))` (the middleware chain expects a synchronous middleware function, not a `Promise<fn>`), and requiring `passedInitObj` breaks production (which intentionally relies on the cached init). 

I hardened it the safe way instead: keep the factory synchronous, but resolve init **lazily inside the middleware** — fast-path when already resolved, otherwise `await` the memoized `getInitObject()` promise rather than rejecting. This removes the fragile fire-and-forget timing dependency without changing the signature. (`807a8b4c`)

**2. Nitpick: redundant `platformMeta` existence check.**
Keeping it — it's required for TypeScript narrowing: `initObj.platforms.get(platformId)` is `T | undefined`, and the earlier `.has()` check doesn't narrow the later `.get()`. Removing the guard would force a non-null assertion. It's a cheap defensive check that also satisfies the type checker.

Server suite 118 pass, lint + build clean.

Issue #1118, opened 2026-06-09
bun server intermittently drops a connection's inbound channel under XMPP-active load (engine.io long-polling)


silverbucket:
## Summary

When the Sockethub **server runs under bun**, a freshly-connected socket can end up with a **working server→client path but a dead client→server (inbound) path**: the server logs the connection, but **no inbound packet from that socket ever reaches socket.io** (no `onAny`, no middleware), so events like `credentials` are never handled and their acks never return. Under **node** the same scenario always works.

This is the root of the intermittent `Test NPM Package` (bun) browser failures (`Timeout waiting for ack … event=credentials`).

## Evidence (from this investigation)

- **node: 100% pass; bun: fails.** Same browser client, same code — only the server runtime differs.
- Server-side tracing (`onAny` + middleware) showed **34 of 35 connections had a working inbound channel; exactly one did not** — the failing socket appeared in zero inbound traces, only `connection` + `disconnect`.
- The lost packet is **below socket.io**, at the **engine.io HTTP long-polling** layer (confirmed via engine.io's own debug logs; `maxPayload: 1000000`).
- `ready()` still resolved because the server **pushes** schemas unsolicited (server→client works); only inbound is dead.

## What it is NOT (ruled out with isolated repros)

- Not a test flaw (the test simulates a legitimate browser refresh/reconnect).
- Not the transport upgrade or transport type (polling-only reproduces it; ws-only fails to connect for unrelated reasons).
- Not generic bun message-dropping: minimal bun + socket.io with connection churn, a broadcast storm, rate-limit middleware, heavy child-process IPC, a real browser client, **and even the real Sockethub server + Redis run locally** all pass. It only manifests with the **full XMPP-active workload** (real prosody/MUC traffic → large schemas pushes + big concatenated long-polling batches, see #1116 / #1117), which can't be run locally without docker.

## Impact

A reconnecting client against a **bun**-hosted server can intermittently get a dead inbound channel (its events silently never reach the server). 

## Recommendation / next steps

- Treat **node as the supported server runtime** until resolved; consider running the npm-package browser suite under node and keeping bun as a non-blocking signal.
- Reduce long-polling payload pressure (#1116, #1117) — may avoid the trigger.
- Investigate / bump engine.io; produce a minimal upstream repro (the trigger appears to require large/batched long-polling payloads under bun) and file with bun/engine.io.

## Context

Full investigation under PR #1108 lineage; see also #1116 and #1117.

[...]

PR #1127, opened 2026-06-11
ci: make bun leg of Test NPM Package non-blocking (node gating) (#1118)

silverbucket:
## Why
The `Test NPM Package` workflow has been red on master since ~2026-01-23 because its **bun** `browser-multiclient` leg fails with `Timeout waiting for ack … event=credentials`. Investigation (see #1118) established:

- **Not a Sockethub code bug, not a payload-size issue** (dropped packet is 393 B; registry push ~4.8 KB).
- **Not an engine.io/socket.io version issue per se** (node passes with the same deps).
- **Not a bun-version regression** — reproduces on **bun 1.3.6**, the exact last-known-good version from the Jan-23 run.
- It's a **bun runtime** interaction with engine.io HTTP long-polling on a **loopback** socket under XMPP-active load: **node passes**, and the **docker**-based Integration workflow (which runs the same multiclient test) **masks** it.

Every attempt to isolate it further via CI (downgrade deps, check out old code, force websocket transport) introduced a *different* confound, so it isn't root-causable by version-pinning. The real fix needs an upstream bun/engine.io repro.

## What this does
Makes **node the gating runtime** and the **bun leg `continue-on-error`**, so the workflow reflects the supported-runtime status (green) while keeping the bun result **visible but non-blocking**. No production code changes.

## Follow-up
#1118 retains the full investigation and tracks the upstream bun-loopback-long-polling issue. The eventual fix lives there (or in a bun release), not here.

## Summary by CodeRabbit

* **Chores**
  * Updated test workflow configuration to improve reliability of testing process.

### DIFF
--- a/.github/workflows/test-npm-package.yml
+++ b/.github/workflows/test-npm-package.yml
@@ -44,6 +44,15 @@ jobs:
   test-packaged-current:
     runs-on: ubuntu-latest
 
+    # node is the gating runtime. The bun leg is non-blocking: it surfaces a
+    # known bun-runtime issue with engine.io HTTP long-polling on a loopback
+    # socket under XMPP-active load (browser-multiclient "Timeout waiting for
+    # ack") that node does not hit and docker masks. It is NOT a Sockethub code
+    # bug and NOT a bun-version regression (reproduces on the last-known-good
+    # bun 1.3.6). Tracked in #1118; keep it visible but non-blocking until the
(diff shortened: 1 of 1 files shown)


silverbucket:
Superseded by #1128, which removes the bun leg from Test NPM Package entirely (node is now the only supported deployment runtime). The non-blocking workaround here is no longer needed.

Issue #918, opened 2025-09-22
Cannot start server: "A Proxy's target should be an Object"


raucao:
I tried to start Sockethub from current master as well as a couple of the open PR branches. The server crashes with the following error message:

<img width="1659" height="486" alt="Image" src="https://github.com/user-attachments/assets/4c0728d1-2687-482c-9667-3283aec1d3d2" />

silverbucket:
@raucao the issue was in sentry, updating to the latest fixes it. finishing the PR now

PR #1128, opened 2026-06-11
chore: Node.js is the deployment runtime; Bun is dev-only tooling

silverbucket:
Makes **Node.js the official, only-supported deployment runtime** and demotes **Bun to a development/build tool** (package manager, builder, unit-test runner). Per the assessment, Sockethub already ran node-compatibly — **zero `Bun.*` runtime APIs, node-target ESM builds, ioredis** — so this is a packaging/identity/CI change, **not a runtime port**. **No production dependencies change.** Unit tests stay on `bun:test` (kept as recommended — a dev/logic concern).

## Runtime
- **Dockerfile** — prod stage now `FROM node:22-slim` running `node …/bin/sockethub`. Build stage keeps Bun (build tool emitting node output); `node:slim` is Debian/glibc-matched to the build image so native-module ABI stays valid.
- **bin shebangs** — `packages/{sockethub,server}/bin/sockethub` → `node` (plain JS, already node-run by CI). `sockethub-logged` stays Bun (it's a TS, `import.meta.path` dev launcher) and is now exposed as `start:logged`.
- **engines** — every package declares `"node": ">=22"` instead of `"bun"`.
- **Removed** root `preinstall: only-allow bun` (it hard-blocked `npm install`).
- **start scripts** → node; build/dev/doc/test scripts stay Bun.

## CI / release validation → node
- `test-npm-package.yml` matrix is **node-only** — releases are validated on the deployment runtime. (Supersedes #1127, which made the bun leg non-blocking; the bun leg is now removed.)
- Docker-based integration/stress jobs now run the server on **node** via the Dockerfile change. `bun install`/`bun run build`/`bun test` steps remain (dev tooling).

## Docs
README, getting-started, server README, and AGENTS now present **Node.js as the runtime/deployment target** and clearly mark **Bun as the dev toolchain**. Contributing/dev sections keep Bun.

## What deliberately did NOT change
- Production dependencies (already node-native).
- Unit tests (bun:test, run with bun) — kept as-is.
- Bun as the build/install/dev tool.

## Bonus
Because production no longer runs on bun, this **sidesteps #1118** (the bun-loopback engine.io long-polling bug) for deployments.

Validated locally: full build green, all package.json valid, unit suites green (schemas 62 / server 118 representative), lint + markdownlint clean. CI (node) validates the deployment paths.

## Summary by CodeRabbit

* **Documentation**
  * Updated install/deploy docs and READMEs to target Node.js v20+ (Bun noted for development/tooling).

* **Chores**
  * Switched CI workflow to test Node.js only.
  * Updated Docker production image and startup to use Node.js.
  * Updated package engine constraints and startup scripts across packages to require Node.js.

### DIFF
--- a/.github/workflows/test-npm-package.yml
+++ b/.github/workflows/test-npm-package.yml
@@ -75,9 +75,8 @@ jobs:
           INPUT_RUNTIME: ${{ inputs.runtime || github.event.inputs.runtime || 'node' }}
           MATRIX_RUNTIME: ${{ matrix.runtime }}
         run: |
-          if [ "$INPUT_RUNTIME" = "both" ]; then
-            echo "run_test=true" >> "$GITHUB_OUTPUT"
-          elif [ "$INPUT_RUNTIME" = "$MATRIX_RUNTIME" ]; then
+          # node is the only matrix runtime; run when it matches (default 'node').
+          if [ "$INPUT_RUNTIME" = "$MATRIX_RUNTIME" ]; then
             echo "run_test=true" >> "$GITHUB_OUTPUT"
           else
             echo "run_test=false" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/test-npm-package.yml
+++ b/.github/workflows/test-npm-package.yml
@@ -8,9 +8,9 @@ on:
         required: false
         default: ''
       runtime:
-        description: Runtime to test (bun, node, both)
+        description: Deployment runtime to test (node)
         required: false
-        default: both
+        default: node
       suite:
         description: Test suite to run (all, browser)
         required: false
@@ -23,10 +23,10 @@ on:
         type: string
         default: ''
       runtime:
-        description: Runtime to test (bun, node, both)
+        description: Deployment runtime to test (node)
         required: false
         type: string
-        default: both
+        default: node
       suite:
         description: Test suite to run (all, browser)
         required: false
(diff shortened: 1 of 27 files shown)


silverbucket:
Addressed all review feedback (incl. nitpick + outside-diff):

1. **[outside-diff] `@sentry/bun` → `@sentry/node`** — good catch. The server runs on Node now, so it should use the Node Sentry SDK. Swapped the dependency and the `sentry.ts` import; updated `sentry.test.ts` (the #918 Proxy bug was specific to `@sentry/bun`'s Bun-serve instrumentation, which `@sentry/node` doesn't have). `sentry.init`/`captureException` are identical across the SDKs. Server suite green (118).

2. **[nitpick] `test-npm-package.yml` runtime logic** — removed the obsolete `both` branch; it now simply runs when `INPUT_RUNTIME` matches the (node-only) matrix.

3. **Node version mismatch (AGENTS / getting-started "≥20" vs packages)** — already resolved, but the *other* way than suggested: I lowered `engines.node` to **>=20** across all packages (CI and the LTS floor are Node 20; `>=22` produced EBADENGINE on the node-20 runner and actually broke the Test NPM Package node run). Docs and packages now both say **>=20**.

4. **[nitpick] Bun toolchain version** — getting-started `v1.2+` → **v1.2.4+** to match the engines floor / AGENTS.

Validated: build + full server suite + lint green, and the node leg of `Test NPM Package` now passes end-to-end on this branch.

PR #1129, opened 2026-06-11
feat(server): dedup platform registry pushes by fingerprint (#1117)

### LINKED ISSUE #1117: Full platform schemas registry is pushed to every client on connect (large payload, esp. over long-polling)
## Summary

On every connection the server emits the **entire platform schemas registry** unsolicited (`socket.emit("schemas", platformRegistryPayload)` in `packages/server/src/sockethub.ts`), and the client also re-requests it during `ready()`. That payload includes the full JSON schemas for **all** platforms (dummy, feeds, irc, metadata, xmpp) and is sizable (tens of KB).

## Evidence

engine.io debug logs show the schemas payload written to each client on connect; with the default **HTTP long-polling** transport and `maxPayload: 1000000` (1MB), these large pushes are concatenated with other buffered messages into single polling responses.

## Impact

- Large per-connection payloads over long-polling.
- Combined with the dead-socket broadcast bloat (#1116), pushes polling batch sizes up toward `maxPayload` — relevant to the bun/engine.io dropped-inbound investigation and to prior reports of >1MB message sizes.

## Proposal

Evaluate reducing payload size/frequency, e.g.: push a compact registry (ids/versions/contexts) and let clients fetch full schemas on demand; and/or avoid re-pushing identical registries (dedupe by fingerprint, which the client already computes).

## Context

Surfaced during the bun npm-package browser-test investigation (PR #1108 lineage).

silverbucket:
Closes #1117.

## Problem
On every connection the server emitted the **entire** platform schemas registry (full JSON schemas for all platforms — tens of KB) unconditionally via `socket.emit("schemas", …)`, and the client *also* re-requested it during `ready()`. So each connection transferred the full schema set at least twice, and every reconnect re-sent it again. Over HTTP long-polling these large pushes concatenate toward `maxPayload`.

## Approach
The client genuinely needs the schemas (it registers them locally and validates outgoing messages with `validateActivity`), so we can't simply stop sending them. Instead we **dedup by fingerprint** — the issue's secondary proposal — and make delivery **on-demand**:

**Server**
- Compute the registry payload + a SHA-256 content fingerprint **once** and cache it (the registry is static after platform load).
- Drop the unconditional on-connect push; serve the registry only in response to the client's `"schemas"` request.
- When the client echoes a fingerprint that matches, reply `{ fingerprint, unchanged: true }` instead of the full payload.

**Client**
- Echo the last-applied fingerprint on each `"schemas"` request.
- Store the **server-supplied** fingerprint (so the echo compares like-for-like).
- On an `unchanged` reply, reuse the cached registry and just complete the init cycle. Falls back to a locally computed fingerprint for payloads without one (e.g. test fixtures), so older payloads still work.

## Result
- **Fresh client / browser refresh:** full registry once (no prior fingerprint).
- **Transport-level reconnect (same client instance):** a tiny `unchanged` reply instead of the full schema set.
- No more per-connection double-send (proactive push + ready() re-request).

`SockethubClient` already requests schemas on every connect/reconnect (`startInitialization → requestSchemaRegistry`), so dropping the proactive push is transparent to it. (Note: a raw, non-`SockethubClient` socket that relied on the unsolicited push must now request it — a minor protocol change.)

## Tests
- New client unit test: echoes the stored fingerprint on reconnect and reuses the cached registry on an `unchanged` reply.
- Updated existing client mocks to find the ack callback among args (the request now carries a fingerprint arg before the callback).
- Full-payload path (fresh connect) is exercised by every integration/browser test.

Local: build + types green, biome clean, client/server/schemas suites 192/192 pass.

## Summary by CodeRabbit

* **New Features**
  * Registry fingerprinting and caching to optimize schema delivery; client echoes fingerprint on reconnect for an unchanged-schema fast path.

* **Bug Fixes**
  * Fixes for fingerprint/unchanged handling so reconnects reuse cached registries without invalidating contexts (regression `#1117`).
  * Client ignores premature/malformed "unchanged" replies when no registry is cached.

* **Tests**
  * Added/updated tests for fingerprint echoing, reconnect registry reuse, and event ack argument handling.

* **Documentation**
  * Clarified client initialization, registry handshake, validation, and reconnect behavior.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -129,7 +129,7 @@ using ActivityStreams as the uniform message format.
 
 - **TypeScript**: Use proper types from `@sockethub/schemas`. Avoid `any`.
 - **Linting**: Biome handles formatting and style. Don't comment on style.
-- **Testing**: Test files named `*.test.ts`. Use `describe` and `test` (not `it`).
+- **Testing**: Test files named `*.test.ts`. Use `describe` and `it` (not `test`).
   Mock external dependencies. Test error paths, not just happy paths.
 - **Error Handling**: Always propagate errors to the caller. Never silently swallow errors.
   Log errors with context (session ID, platform, action).
--- a/docs/client-guide.md
+++ b/docs/client-guide.md
@@ -296,11 +296,14 @@ sc.socket.emit('message', {
 ## Client Features
 
 - **Schema-driven init**: `ready()` resolves when the server's schema registry is loaded
+- **Registry caching**: the registry is fetched once and cached; on reconnect the
+  client echoes a fingerprint so the server skips re-sending an unchanged registry
 - **Context composition**: `contextFor(platform)` builds canonical `@context` arrays
 - **Outbound queueing**: Messages sent before `ready()` are queued and flushed automatically
 - **Auto-replay**: Credentials and connections restored on reconnect
-- **ActivityStream helpers**: Normalization and property linting via `@sockethub/schemas`
-  (can be called explicitly, but also automatically by the client on send/receive)
+- **ActivityStream handling**: outbound messages are normalized and validated against
+  the platform's schema (via `@sockethub/schemas`) before sending; incoming messages
+  are normalized before delivery
 - **Connection state**: Check `sc.socket.connected` for status
 
 ## Reference
--- a/packages/client/README.md
+++ b/packages/client/README.md
@@ -8,12 +8,40 @@ reconnection, and credential replay.
 ## What's Included
 
 - `SockethubClient` for connection and message handling
-- Schema-driven ActivityStreams validation and property linting (via `@sockethub/schemas`)
+- Schema-driven validation of outbound ActivityStreams messages (via `@sockethub/schemas`)
 - `contextFor(platform)` builds canonical `@context` arrays from server metadata
 - `ready()` promise and `ready`/`init_error` observability events
 - Automatic outbound queuing until initialization is complete
 - Auto-replay of credentials and connections on reconnect
 - A browser-ready bundle in `dist/`
+
+## How It Works
+
+`SockethubClient` wraps a [Socket.IO](https://socket.io) connection and manages
+everything between your app and the Sockethub server:
+
+1. **Schema registry handshake.** On connect (and every reconnect) the client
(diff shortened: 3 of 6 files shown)


silverbucket:
Thanks @coderabbitai — addressed in 35382a88.

**1. Fallback dedup could miss real schema changes (Major) — fixed.**
You're right: `computePayloadFingerprint()` only hashed version/contexts/platform metadata, so a schema-body change with an unchanged `schemaVersion` could be skipped. Rather than make the fallback hash the full payload, I dropped the fallback entirely: dedup now keys **only** on the server's fingerprint, which is computed over the full payload (including schema bodies and `types`). When the server sends no fingerprint, `registryFingerprint` is left unset so `handleSchemasPayload` never short-circuits and always re-applies — a schema change can't be silently missed. Removed the now-unused `computePayloadFingerprint` helper.

**2. Validate the reply before honoring `unchanged`/`fingerprint` (Major) — fixed.**
The `unchanged` fast-path now only fires when a registry is actually cached (`registryFingerprint` set **and** `platformRegistry.size > 0`). A malformed or premature `unchanged` reply falls through to `applyPlatformRegistry()`, which rejects the contentless payload and raises an `init_error` — so it fails loudly instead of fast-pathing to ready with no validators registered. The full-payload path was already structurally validated in `applyPlatformRegistry()` (rejects anything without valid `contexts`/`platforms`). I did not add a new AJV schema here: this is server→client data (not untrusted client input), and the existing structural validation + the cache guard cover the failure modes you flagged.

**3. Add a negative test for the `unchanged` path (Nitpick) — added.**
New test: an `unchanged` reply received with no cached registry must not mark the client ready (`isSchemasReady()` stays false and `contextFor()` throws).

**4. `it()` → `test()` (Nitpick) — declined, with reason.**
`packages/client` tests run under **web-test-runner + mocha** (`test:browser`; the file imports chai's `expect` and has no `bun:test` import). Mocha exposes `describe`/`it`, **not** `test` — renaming would break the browser test run. The `describe`/`test` convention applies to the `bun:test` suites (server/schemas), not these mocha browser tests.

Also refreshed the client docs: corrected stale "property linting" wording (the client lint was retired in #1126 — outbound messages are validated, not linted) and documented the schema-registry handshake + fingerprint dedup.

PR #1130, opened 2026-06-11
docs: clarify client/server responsibilities, validation & replay

silverbucket:
## Why
The existing `docs/architecture.md` explained the server↔platform layer well but said nothing about the **client SDK's role** or the **validation model** — which is a recurring source of confusion: *"I thought validation was server-only, so why does the server publish schemas to every client?"*

## What
Adds the missing pieces (with **Mermaid** diagrams, so they render graphically on GitHub/GitBook), inserted after the System Overview; the existing server↔platform / security / scaling sections are unchanged.

- **Component Responsibilities** table — who owns authoritative validation, normalization, schemas, `@context` composition, outbound queueing, reconnect replay, process isolation, and credential storage.
- **Validation Model** — the server is the *sole authoritative* validator, in **both** directions; the client's outbound check is an **advisory fail-fast**, not a trust boundary. Explains the three per-platform schemas and why only the two **inbound** ones (`credentials` + `messages`) are pushed to clients, while `responses` stays server-only:

  | Schema | Direction | Server | Pushed to client | Client |
  |---|---|---|---|---|
  | `credentials` | app→server | ✅ auth | ✅ | ⚠️ advisory |
  | `messages` | app→server | ✅ auth | ✅ | ⚠️ advisory + `contextFor` |
  | `responses` | platform→app | ✅ auth | ❌ | — |

- **Client⇄Server coordination** — sequence diagrams for the connect/schema handshake (including the #1117 fingerprint dedup), the outbound command flow, and the inbound event flow.
- **Reconnection & Replay** — what the client remembers (`credentials`/`connect`/`join` maps), the dependency-ordered replay on reconnect (credentials → connect → join → queued), and that the server re-validates every replayed activity.

## Notes
- Doc-only change; no code touched.
- Describes the schema-handshake fingerprint dedup landing in #1129. Best merged after (or alongside) #1129; the rest of the doc is independent of it.

## Summary by CodeRabbit

* **Documentation**
  * Expanded architecture docs with a clear validation model and component responsibilities.
  * Detailed client↔server coordination flow with sequence diagrams for connect, command send, and inbound events.
  * Added reconnection & replay behavior and ordering for credentials/connect/join/queued messages.
  * Clarified session isolation and encryption/security characteristics of transport, storage, and queued payloads.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -346,9 +346,18 @@ Redis and BullMQ provide reliable, encrypted communication between server and pl
 
 ### Validation Pipeline
 
+The per-session middleware chain runs on each inbound event; the validated job is
+then handed to the job-queue layer, which encrypts the payload as it enqueues:
+
 ```
-Incoming activity → normalize → schema validation → credential handling → encrypt → enqueue
+middleware:  normalize → schema validation → credential handling
+job queue:   queue.add() → encrypt payload → enqueue (Redis/BullMQ)
 ```
+
+Encryption is not a middleware step — it happens inside `@sockethub/data-layer`
+(`JobQueue.createJob` → `encryptActivityStream`) when `platformInstance.queue.add()`
+is called; the worker decrypts via `decryptJobData` before running the platform
+handler.
 
 ## Platform Types
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -42,8 +42,9 @@ Sockethub is a protocol gateway that runs as four coordinated components:
 └────────────────┘     └─────────────────────┘    └─────────────────┘
 ```
 
-**Data Flow**: All ActivityStreams messages, credentials, and job data are encrypted before
-storage or transmission. Each phase handles specific responsibilities:
+**Data Flow**: Credentials and job payloads are encrypted at rest in Redis and on the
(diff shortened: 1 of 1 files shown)


silverbucket:
Thanks @coderabbitai — both assessed.

**1. Clarify where encryption happens (Minor) — fixed.**
You're right that `encrypt` is not a middleware stage. Reworked the Validation Pipeline section to separate the per-session middleware chain (normalize → schema validation → credential handling) from the job-queue layer, and added a note that encryption happens inside `@sockethub/data-layer` (`JobQueue.createJob` → `encryptActivityStream`) when `platformInstance.queue.add()` is called, with the worker decrypting via `decryptJobData`.

**2. Handshake diagram vs implementation (Major) — resolved by merging master.**
The diagram was correct; the mismatch was that this branch was created before #1129 merged, so it didn't yet contain the server-side `{unchanged: true}` reply / client fingerprint-echo you were looking for. #1129 is now merged, and I've merged master into this branch — `packages/server/src/sockethub.ts` now has the `clientFingerprint === fingerprint ? { fingerprint, unchanged: true } : payload` logic, and the client echoes the server-supplied fingerprint, exactly as the diagram shows. So the docs match the in-tree behavior.

PR #1135, opened 2026-06-12
refactor(platform-feeds): remove dead feedparser-era mapping

silverbucket:
## Problem

`podparse`'s `Episode` type (`node_modules/podparse/index.d.ts`) has **no** `meta`, `date`, `categories`, `media`, or `source` fields. The local `FeedItem` interface in `src/index.ts` re-added them via a cast, so in `buildFeedItem` those fields were **always undefined** on real parsed feeds. This was verified against podparse's runtime source as well, not just the typings: episodes are built exclusively from the `rssElements` mapping table, which contains no `meta`/`date`/`categories`/`media`/`source` keys (`meta` exists only on the top-level `Podcast` return, never on episodes). As a result:

- `buildFeedItem` emitted `updated`, `tags`, `media`, and `source` from fields that never exist on a parsed `Episode`.
- The `datesEqual` helper (and its `updated` logic) only ever exercised hand-built test objects, never real feeds.
- `PlatformFeedsActivityObject` marked every field as required while the code assigned `undefined` to several of them — the interface lied about the shape.
- The outbound `responses` schema declared optional `updated`/`tags`/`media`/`source` properties the code never emits.

## Fix

- `buildFeedItem` now takes `Episode` directly (the `FeedItem` cast is removed). It drops the `item.meta?.link` fallback (uses `item.link` then `channelUrl`) and no longer emits `updated`, `tags`, `media`, or `source`. It keeps `title`, `id`/stableId, `type`, `brief`, `content`, `contentType`, `url`, `published`, `datenum`.
- Removed the `datesEqual` export and its only usage, plus its dedicated `describe` block and the dates-related cases in `index.test.ts`.
- Trimmed `PlatformFeedsActivityObject` to remove `updated`/`tags`/`media`/`source` and made `brief?`/`published?` optional to match what the code actually emits.
- Removed the now-dead optional `updated`/`tags`/`media`/`source` properties from the `feedItem` schema so the schema honestly describes the emitted shape. (`additionalProperties: false` is unchanged, so strictness is preserved.)
- The fixture-matrix tests now pass real podparse episodes to `buildFeedItem` without a cast, so the compiler verifies the new `Episode`-typed signature against actual podparse output.

This is **behavior-preserving**: `JSON.stringify` already dropped the always-`undefined` keys from the wire output, so removing them changes nothing on the wire.

Note: the removed `tags: item.categories` was dead because the field name was wrong, not because the data doesn't exist — podparse exposes real RSS `<category>` tags as `episode.category` (singular, untyped on `Episode`). A future feature could re-add tags by reading `item.category` and re-declaring it in the schema.

## Rebase onto master (post-#1132 / #1134)

This branch has been rebased onto current master, which already contains #1132 (string channel `image`/`author` handling, post `id` in the schema) and #1134 (strict fetch object schema with `since`/`limit` filtering). Conflict resolution:

- `src/index.ts`: kept master's `FeedFetchParams`, `extractFetchParams`, and `applyFetchFilters`; deleted only the `FeedItem` interface and `datesEqual`. (`applyFetchFilters` filters on `object.datenum`, which this PR preserves — including the `datenum === 0` unparseable-date sentinel — so `since`/`limit` filtering is unaffected.)
- `src/index.test.ts`: kept master's import list minus `datesEqual` and kept #1134's `inbound messages schema` suite; deleted only the `datesEqual` suite. The two rewritten `buildFeedItem` tests now use `it()` per current test convention.
- `schema.ts`/`types.ts` auto-merged cleanly: #1132's post `id` and string `image`/`author` handling are intact.

No further rebase dependency remains: the previously anticipated `datesEqual` import conflict with #1134 was resolved in this rebase. Agreed merge order for the remaining open work: this PR (#1135) before #1133.

## Verification

- `bun test` in `packages/platform-feeds`: 37 pass, 0 fail.
- `bun run lint` (repo root): clean.
- `bun run build` in the package: succeeds.
- `bunx tsc --noEmit`: identical error count to master (all pre-existing, none in platform-feeds).

## Summary by CodeRabbit

* **Refactor**
  * Feed item structure simplified: removed optional fields (updated, tags, media, source); brief and published are now optional.
  * Feed builder now consumes raw episode input and derives published/datenum from pubDate.

* **Tests**
  * Tests reorganized to async/await, simplified fixtures and fuzzing.
  * Added URL/id fallback test; removed updated-date comparison tests and related helper.


silverbucket:
Re the outside-diff finding on `buildFeedItem` ("Omit `published` when `pubDate` is missing"): verified and not an issue in this codebase.

- The validator is AJV (`packages/schemas/src/validator.ts`), and AJV treats an own property whose value is `undefined` as absent — verified empirically with the project's AJV options against this exact strict `feedItem` schema: `{ published: undefined }` validates clean.
- The only runtime response validation (`packages/server/src/platform-instance.ts` `sendToClient`) happens after the platform child-process IPC boundary, where Node's default JSON serialization has already dropped `undefined`-valued properties.
- The in-repo fixture tests that validate raw `buildFeedItem` output against the strict outbound schema also pass.

So `published: undefined` cannot fail schema validation in any path this codebase has. A conditional spread would be cosmetic only, and would be inconsistent with the adjacent `brief: ... undefined` pattern on the line above; no change made.

silverbucket:
Addressed in eb1d6934. Converted the remaining callback-style `platform.fetch` e2e tests ("fetches expected feed", "handles empty feed gracefully", "handles feed items without per-entry link", "handles regression podcast feed without per-entry link or meta", "handles feed with no actor name", "validates collection structure...") to await the existing `fetchCollection` helper, and the "handles network errors properly" case now asserts via `await expect(...).rejects.toThrow("Network timeout")`. These are now `it()` per convention. Verified a deliberately broken assertion fails the suite (it silently passed before). 37/37 pass, lint/build green.

PR #1141, opened 2026-06-14
fix(platform-irc): never send bare NAMES on attendance query

### LINKED ISSUE #1085: IRC: User presence messages from unjoined channels
While testing unrelated things in Hyperchannel with recent Sockethub PRs, I just received a ton of presence messages from rooms I had never joined. Not sure if it's a Hyperchannel integration issue or Sockethub issue, but it has to be looked into before a stable release.

silverbucket:
## Summary

Fixes the IRC presence flood in #1085. An attendance `query` whose target carried only an `id` (no `name`) caused the platform to emit a **bare `NAMES`** command, which IRC servers answer with the entire network channel list — flooding the client with presence for rooms it never joined.

`query()` now resolves the channel from `target.name`, falling back to the channel segment of `target.id` (`<server>/<channel>`), and **rejects the query when no valid `#channel` can be determined**, so a bare `NAMES` is never sent.

## How it happened

- `query()` built the command as `client.raw(["NAMES", job.target.name])`.
- Clients that send `target: { id, type: "room" }` without `name` (e.g. Hyperchannel's `queryAttendance`, see 67P/hyperchannel#312) leave `target.name` undefined.
- `["NAMES", undefined].join(" ")` → `"NAMES "` → bare NAMES → server returns `353` for every visible channel → `irc2as` emits a presence per user per channel → all forwarded to the client.

Deriving from `target.id` means the flood is prevented even for clients that only send an id.

## Changes

- `packages/platform-irc/src/index.ts`: add `resolveChannelName()`; guard `query()` against an unresolvable channel.
- `packages/platform-irc/src/index.test.ts`: regression coverage — NAMES from `target.name`, channel derived from `target.id`, and rejection without a bare NAMES when no channel resolves.

## Verification

- `bun test packages/platform-irc/` → 84 pass / 0 fail
- `bun run build` (platform-irc) succeeds; biome clean

## Notes

Defense-in-depth filtering of inbound presence by joined channels was considered but deferred: with the bare-NAMES source closed and nothing in the stack issuing `WHO`, its marginal value is low while it risks dropping the legitimate initial `353` that arrives after a `JOIN` (membership is recorded only on the subsequent `PING`/`PONG`). Can follow up if wanted.

Closes #1085

## Summary by CodeRabbit

* **Tests**
  * Expanded test coverage for attendance query functionality, including channel name resolution scenarios and error cases.

* **Bug Fixes**
  * Improved channel name resolution for attendance queries with better error handling when valid channels cannot be determined.
  * Added logging for resolved channel names to aid in debugging.

### DIFF
--- a/packages/platform-irc/src/index.test.ts
+++ b/packages/platform-irc/src/index.test.ts
@@ -500,18 +500,74 @@ describe("Initialize IRC Platform", () => {
                 platform.completeJob();
             });
 
-            it("query()", (done) => {
-                platform.query(
-                    {
-                        "@context": IRC_CONTEXT,
-                        type: "query",
-                        actor: actor,
-                        target: targetRoom,
-                        object: { type: "attendance" },
-                    },
-                    done,
-                );
-                platform.completeJob();
+            describe("query() attendance", () => {
+                let rawCalls;
+                beforeEach(() => {
+                    rawCalls = [];
+                    platform.client.raw = (...args) => {
+                        rawCalls.push(args);
+                    };
+                });
+
+                it("sends NAMES for the target channel name", (done) => {
+                    platform.query(
+                        {
+                            "@context": IRC_CONTEXT,
+                            type: "query",
+                            actor: actor,
+                            target: targetRoom,
+                            object: { type: "attendance" },
+                        },
+                        (err) => {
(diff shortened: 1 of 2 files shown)


raucao:
Nice!

PR #1143, opened 2026-06-17
fix(platform-irc): explicit TLS cert validation + working self-signed opt-out

### LINKED ISSUE #1056: security(platform-irc): TLS certificate validation is disabled (rejectUnauthorized: false)
## Summary

`packages/platform-irc/src/index.ts` sets `rejectUnauthorized: false` on the TLS socket unconditionally whenever `secure: true`:

```ts
if (is_secure) {
    module_options.connectOptions = { rejectUnauthorized: false };
}
```

This means users connecting over TLS are **not** getting certificate authenticity — only encryption. A network-path attacker (malicious WiFi, ISP, state-level actor, etc.) can present any certificate and MITM the connection, harvesting the SASL secret (password or OAuth token) in transit.

## Impact

* Any SASL credential — password **or** token — sent over a "secure" IRC connection is exposed to a trivial MITM on the network path.
* With the recent addition of SASL OAUTHBEARER support (#1055), this is arguably worse, since OAuth tokens can carry broader and longer-lived access than a single IRC-network password.

## Proposed fix

Switch the default to validating certificates, and provide an explicit per-credentials opt-out for users connecting to IRC networks with self-signed certificates:

```ts
// schema (credentials.object)
allowInvalidCert?: boolean  // default: false
```

```ts
// index.ts
if (is_secure) {
    module_options.connectOptions = {
        rejectUnauthorized: !credentials.object.allowInvalidCert,
    };
}
```

## Acceptance criteria

- [ ] Default behaviour validates TLS certificates (`rejectUnauthorized: true`)
- [ ] New opt-in credential field lets users intentionally allow invalid certs
- [ ] Credential schema updated + tests
- [ ] Documented in `packages/platform-irc/README.md` Authentication section
- [ ] Note the default change in the release notes (behaviour change for anyone currently relying on invalid certs)

## Context

Surfaced during security review of #1055 (OAUTHBEARER token auth). Kept out of that PR to keep scope focused.

mvanhorn:
## Summary

`ircConnect()` in `packages/platform-irc/src/index.ts` set `rejectUnauthorized: false` unconditionally for every secure (`secure: true`) IRC connection, so TLS sessions were encrypted but never authenticated. This change validates the server's TLS certificate by default and adds an explicit, opt-in escape hatch for self-signed networks:

- `rejectUnauthorized` is now derived from the credentials: `!credentials.object.allowInvalidCert`, defaulting to certificate validation ON.
- New optional `allowInvalidCert?: boolean` credentials field (default `false`), declared in the JSON schema (`schema.ts`), the `PlatformIrcCredentialsObject` interface (`types.ts`), and documented in the README Authentication section.
- A thin TLS transport wrapper flattens the connect options before they reach `tls.connect`. `irc-socket-sasl` builds the options object it hands to the transport with `Object.create(connectOptions)`, which places `rejectUnauthorized` on the prototype rather than as an own property; Node's `tls.connect` ignores it there, so without this wrapper neither the secure default nor the opt-out would actually take effect.

## Why this matters

Without certificate validation, a network-path attacker can present any certificate and transparently MITM a "secure" IRC connection, harvesting the SASL password or OAuth bearer token in transit (#1056, surfaced during security review of the OAUTHBEARER work). Encryption without authentication provides no protection against an active attacker. This restores the security guarantee secure connections are expected to carry, while preserving compatibility with self-signed IRC networks through an explicit, documented opt-out.

This is a behaviour change: deployments that relied on connecting to IRC servers with invalid/self-signed certificates over `secure: true` must now set `allowInvalidCert: true`. Worth a release-notes callout.

## Testing

- `bun test` in `packages/platform-irc`: 90 pass / 0 fail. New tests cover schema validation of `allowInvalidCert` (boolean accepted, non-boolean rejected), the `rejectUnauthorized` computation for secure-default / opt-out / cleartext paths, and the option-flattening regression (an inherited `rejectUnauthorized` becomes an own property).
- `bun run lint` (Biome + markdownlint) and `bun run build` pass.
- Verified end to end against a live self-signed TLS server: a default secure connection is rejected (`DEPTH_ZERO_SELF_SIGNED_CERT`), and `allowInvalidCert: true` connects successfully.

Fixes #1056

## Summary by CodeRabbit

* **New Features**
  * TLS certificate validation is now enabled by default for secure IRC connections.
  * Added an `allowInvalidCert` credential option to allow trusted self-signed certificates.

* **Documentation**
  * Updated authentication/connection docs to explain TLS certificate validation behavior and the new option.

* **Bug Fixes**
  * Secure connections now correctly apply certificate validation settings, while non-secure connections no longer include TLS options.

* **Tests**
  * Expanded coverage for `allowInvalidCert` (boolean acceptance/rejection) and secure vs non-secure TLS behavior.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -152,7 +152,7 @@
       "version": "4.0.0-alpha.12",
       "dependencies": {
         "@sockethub/irc2as": "^4.0.0-alpha.12",
-        "irc-socket-sasl": "^4.1.1",
+        "irc-socket-sasl": "^4.1.2",
       },
       "devDependencies": {
         "@sockethub/schemas": "^3.0.0-alpha.12",
@@ -2333,7 +2333,7 @@
 
     "ipaddr.js": ["ipaddr.js@1.9.1", "", {}, "sha512-0KI/607xoxSToH7GjN1FfSbLoU0+btTicjsQSWQlh/hZykN8KpmMf7uYwPW3R+akZ6R/w18ZlXSHBYXiYUPO3g=="],
 
-    "irc-socket-sasl": ["irc-socket-sasl@4.1.1", "", { "dependencies": { "r-result": "^1.5.1" } }, "sha512-uqCSRFu61VUfRAhYit6tOgX7Ln2FEJqiGQLYhB0HMEWnWQ5fMcWAFhefggu7Q+PjdLlP5dOPOiw6Z50J5tE0Ew=="],
+    "irc-socket-sasl": ["irc-socket-sasl@4.1.2", "", {}, "sha512-ioBVPtvQE3bBNZAqLF8XnWj4/pSEsGvsrq2j8OErDScc3Xvwf/9kmUFLdOrEY7YqugwPmkloVKH7UQngtkUOlQ=="],
 
     "is-alphabetical": ["is-alphabetical@2.0.1", "", {}, "sha512-FWyyY60MeTNyeSRpkM2Iry0G9hpr7/9kD40mD/cGQEuilcZYS4okz8SN2Q6rLCJ8gbCt6fN+rC+6tMGS99LaxQ=="],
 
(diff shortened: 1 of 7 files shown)


silverbucket:
Thanks for this — merging shortly. Wanted to leave a note correcting the severity framing before it goes in, and record what changed.

**Root cause was upstream, not here, and the current default was already safe.** While reviewing this I traced the actual behavior through `irc-socket-sasl`: it builds `connectOptions` via `Object.create(connectOptions)` (irc-socket.js:87), which puts the caller's `rejectUnauthorized` on the *prototype* rather than as an own property. Node's `tls.connect` merges options internally with an object spread, which only copies own enumerable properties — so the inherited `rejectUnauthorized` was silently dropped either way, and Node fell back to its own default (validate). I verified this end-to-end against a real self-signed server with the actual pre-PR code: it rejects with `DEPTH_ZERO_SELF_SIGNED_CERT` regardless of whether `rejectUnauthorized` was `true` or `false` upstream of the `Object.create` step.

So there wasn't an active MITM exposure on master — the real bug was the mirror image of what's described: **secure connections to self-signed IRC servers were unconditionally broken**, with no way to opt out, because the setting never reached `tls.connect` either way.

That doesn't take away from the fix here — making the default explicit and giving `allowInvalidCert` a working opt-out is exactly right, and the `flattenConnectOptions`/`tlsTransport` wrapper is a correct, well-tested solution to the real underlying issue (verified it end-to-end myself against a self-signed server, both the default-reject and opt-out paths).

**Filed and fixed upstream too:** opened silverbucket/irc-socket-sasl#24 with the root cause and a one-line fix (swap `Object.create` for a plain spread copy), since the same silent-drop affects any `connectOptions` field, not just `rejectUnauthorized`, for every consumer of that library. `4.1.2` shipped with the fix.

I've pushed a follow-up commit bumping `irc-socket-sasl` to `^4.1.2` and updating the comment above `flattenConnectOptions` to reflect that it's now a defensive backstop (in case of a downgrade or a fork reintroducing the pattern) rather than the primary fix — the wrapper itself is unchanged and still needed since we shouldn't depend on the installed dependency version alone to get this right.

mvanhorn:
Thanks @silverbucket! Explicit TLS validation on the IRC platform closes a quiet gap.

mvanhorn:
Thanks for the merge, @silverbucket! Glad the explicit TLS cert validation for the IRC platform is in.

PR #1153, opened 2026-07-04
perf(data-layer): parallel job processing for stateless platforms

silverbucket:
Stateless platforms (feeds, metadata) run as a single shared child
process for every session on the server, and their BullMQ worker used
the default concurrency of 1 — every fetch for every client was
processed serially, so one slow or hanging feed stalled all other
clients' jobs server-wide.

JobWorker now accepts a concurrency option, and the platform child
process sets it from the platform config: persistent platforms stay
serial (their handlers share per-actor connection state), stateless
platforms default to 10 and can be tuned per-platform via
packageConfig.concurrency.

## Summary by CodeRabbit

* **New Features**
  * Added per-platform configurable job-processing concurrency, enabling parallel processing for non-persistent platforms.
  * Persistent platforms continue to process jobs serially.
  * Concurrency can be set through platform configuration (with schema support) and is resolved during worker startup.
* **Bug Fixes**
  * Added validation to ensure concurrency is a finite integer ≥ 1.
  * Updated startup logging to show the resolved concurrency level.

### DIFF
--- a/packages/data-layer/src/index.ts
+++ b/packages/data-layer/src/index.ts
@@ -14,7 +14,7 @@ import {
     resetSharedRedisConnection,
 } from "./job-base.js";
 import { JobQueue, verifyJobQueue } from "./job-queue.js";
-import { JobWorker } from "./job-worker.js";
+import { JobWorker, type JobWorkerOptions } from "./job-worker.js";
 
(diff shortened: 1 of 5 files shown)


silverbucket:
@coderabbitai review

PR #1155, opened 2026-07-04
fix(server): cap concurrent platform instances

silverbucket:
Every persistent-platform actor forks its own Node child process, and
there was no upper bound on how many could be created. On a public
instance this is a resource-exhaustion vector: enough distinct actors
(or a misbehaving client cycling credentials) exhausts memory and
Redis connections.

Adds limits.maxPlatformInstances (default 0 = disabled, preserving
current behavior). When the cap is reached, ProcessManager refuses to
fork a new instance and the client receives the request back with an
error instead of the server crashing on an unhandled throw. Reusing an
existing live instance is always allowed.

## Summary by CodeRabbit

* **New Features**
  * Added a new configuration to cap concurrently running platform instances, with the default set to `0` (limit disabled).
  * The server enforces capacity by blocking creation of new instances once the cap is reached.

* **Bug Fixes**
  * When the instance limit is hit, requests now fail gracefully with a clear error message instead of throwing.
  * Live/reusable instances are reused when possible, including replacing an actor’s own dead instance even at the cap.

* **Tests**
  * Added automated coverage for the platform instance limit behavior.

### DIFF
--- a/packages/schemas/src/schemas/sockethub-config.ts
+++ b/packages/schemas/src/schemas/sockethub-config.ts
@@ -150,6 +150,17 @@ export const SockethubConfigSchema = {
                 },
             },
         },
+        limits: {
+            type: "object",
+            additionalProperties: false,
+            properties: {
+                maxPlatformInstances: {
+                    type: "number",
+                    minimum: 0,
+                    default: 0,
+                },
+            },
+        },
         redis: {
             type: "object",
             additionalProperties: false,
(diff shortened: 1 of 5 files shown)


silverbucket:
@coderabbitai code review

silverbucket:
Both addressed in 395e641d:

1. **Off-by-one at the cap boundary** — `assertInstanceCapacity` now takes the `identifier` and excludes its own existing (dead) slot from the occupied count, so replacing a crashed instance for the same actor no longer trips the cap.
2. **Test coverage** — added `process-manager.test.ts` covering: cap disabled preserves current behavior, a new actor is blocked at the cap, a live instance is always reused regardless of the cap, and the same actor can replace its own dead instance at the cap. Confirmed the new test fails against the pre-fix code and passes after.

PR #1156, opened 2026-07-04
fix(server): collision-free job titles, expire orphaned handlers

silverbucket:
Job titles keyed the per-instance completed-job handler registry but
were built from client-supplied msg.id when present. Reused or
colliding ids meant two in-flight jobs shared a title: the second
registration silently overwrote the first, dropping or misrouting the
first client's response. Titles now always use the queue's internal
counter.

Separately, handlers for jobs that never produce a result (platform
process dies mid-job, queue paused during teardown) were never removed
from completedJobHandlers, growing the map for the lifetime of the
instance. Registration now goes through registerCompletedJobHandler(),
which timestamps entries and prunes anything older than 6 minutes
(the queue's own removeOnComplete/removeOnFail age plus slack).

### DIFF
--- a/packages/data-layer/src/job-queue.test.ts
+++ b/packages/data-layer/src/job-queue.test.ts
@@ -123,16 +123,17 @@ describe("JobQueue", () => {
                 id: "an identifier",
             });
             expect(job).to.eql({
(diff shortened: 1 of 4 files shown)


silverbucket:
@coderabbitai code review

PR #1157, opened 2026-07-04
fix(server): single process listener pair per platform instance

silverbucket:
Each registered session attached its own message and close listener to
the platform child process. On shared instances — including the global
feeds/metadata platforms every session touches — this meant O(sessions)
listener invocations per platform emit, MaxListenersExceededWarning
spam past ten sessions (nothing raised the limit), and listeners that
lingered until a janitor cycle noticed the socket was gone.

The instance now attaches exactly one message and one close listener at
construction and routes to registered sessions itself. Fatal errors
(error IPC messages, unexpected process close) are now broadcast to
every connected session before shutdown; previously only the session
whose listener ran first received the error and everyone else sharing
the instance lost the platform silently.

registerSession() just records the session; the janitor no longer
removes per-session listeners.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stale socket session cleanup by directly clearing stale session and IP references during janitor cleanup.
  * Reworked child-process close/error handling to broadcast normalized error events to all connected sessions, then shut down.
  * Centralized IPC routing with shared process-level listeners for more consistent message delivery and error notifications.

* **Tests**
  * Updated janitor and platform instance test suites to match the new session tracking and broadcast-based fatal error behavior.

### DIFF
--- a/packages/server/src/janitor.test.ts
+++ b/packages/server/src/janitor.test.ts
@@ -21,18 +21,6 @@ function getPlatformInstanceFake() {
             ["session foo", "127.0.0.1"],
             ["session bar", "127.0.0.1"],
         ]),
-        sessionCallbacks: {
-            close: (() =>
-                new Map([
-                    ["session foo", function sessionFooClose() {}],
-                    ["session bar", function sessionBarClose() {}],
-                ]))(),
-            message: (() =>
-                new Map([
-                    ["session foo", function sessionFooMessage() {}],
-                    ["session bar", function sessionBarMessage() {}],
-                ]))(),
-        },
     };
 }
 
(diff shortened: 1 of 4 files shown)


silverbucket:
@coderabbitai code review

silverbucket:
Both CodeRabbit nits addressed in 68587e48:

1. Dropped the vestigial `process.removeListener` stub from the janitor test fake — `janitor.ts` no longer touches `.process` at all post-refactor.
2. `attachProcessListeners`'s message/close callbacks now `.catch` and log rejections from `handleProcessMessage`/`handleProcessClose` instead of discarding them via `void`, matching the repo's IPC-error-handling guideline.

Full server suite (143 tests) and biome both pass.

silverbucket:
Addressed the outside-diff finding in 12d67e72 — confirmed it's real, not a false positive.

I forked a real child process locally and logged `child.connected` at each lifecycle event: `disconnect`, `exit`, and `close` all report `connected: false`. So `handleProcessClose`'s `this.process?.connected && !this.flaggedForTermination` gate was never true on an unexpected close — the branch that broadcasts the fatal error was dead code, and every close (expected or not) silently fell through to `shutdown()` with no client notification. That's exactly the bug this PR's description says it fixes, so this needed a real fix, not just a comment.

Fix: gate on `flaggedForTermination` alone (every intentional termination path — `shutdown()`, credential-init failure, heartbeat timeout — already sets it before tearing anything down, so it alone distinguishes expected vs. unexpected closes). Also moved `shutdown()` outside the branch: previously it only ran in the 'already flagged' branch, so a genuine crash with live sessions attached would leak the instance forever (the janitor only checks `sessions.size`/`isInitialized()`, not process liveness).

Updated `platform-instance.test.ts` accordingly — the old 'skip error reporting when process disconnected' test was asserting the buggy behavior, so I replaced it with one that reflects the real runtime invariant (connected is always false at close) and confirms both the broadcast and the cleanup now happen. Verified the new test fails against the pre-fix code. Full suite (142 tests) and biome pass.

PR #1158, opened 2026-07-04
fix(server): configurable CORS origin, per-IP connection cap

silverbucket:
The socket.io server hardcoded cors.origin '*': any website could
connect its visitors' browsers to a Sockethub instance and use it as a
relay (IRC spam, feed-fetch proxying). The origin is now configurable
via sockethub.cors.origin — '*' (default, preserving behavior), a
single origin, or a comma-separated list. Public deployments should
set an explicit origin.

The per-event rate limiter is keyed by socket id, so a client could
bypass it by opening additional sockets. rateLimiter.maxConnectionsPerIp
(default 0 = disabled) now caps concurrent connections per client IP;
sockets over the cap receive an error message and are disconnected.

### DIFF
--- a/packages/schemas/src/schemas/sockethub-config.ts
+++ b/packages/schemas/src/schemas/sockethub-config.ts
@@ -148,6 +148,11 @@ export const SockethubConfigSchema = {
                     type: "number",
                     default: 5000,
                 },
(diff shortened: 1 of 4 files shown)


silverbucket:
@coderabbitai code review

PR #1164, opened 2026-07-05
ci: restore release package-test gate and packaged-test config

silverbucket:
## Summary

The `test-local-packages` gate in release-publish has been a silent no-op since #1106: it passes `runtime: both`, but the test matrix was reduced to `[node]`, so the exact-match runtime check set `run_test=false`, every step was skipped, and the job reported success. **v5.0.0-alpha.13 was published with zero packaged-test verification** ([evidence](https://github.com/sockethub/sockethub/actions/runs/28725985062) — all steps after "Determine runtime" skipped).

Re-running Test NPM Package against `v5.0.0-alpha.13` ([run](https://github.com/sockethub/sockethub/actions/runs/28738438509)) then surfaced what the gate would have caught: `browser-basic` and `browser-contract` fail because the packaged-test harness spawns sockethub with no config file, so the SSRF-guarded fetch (#1133/#1138/#1140) blocks the loopback test fixtures (`http://localhost:10550/feed.xml`). The in-repo integration jobs don't hit this because docker-compose mounts `integration/sockethub.ci.config.json` (`allowPrivateAddresses: true`) via `SOCKETHUB_CONFIG`.

## Changes

- **release-publish.yml**: `runtime: both` → `runtime: node`, restoring the release gate
- **test-npm-package.yml**: fail loudly on an unrecognized runtime value instead of skipping every step, so a caller mismatch can never again produce a vacuous green job
- **test-npm-package.yml**: normalize a bare-semver `ref` input (`5.0.0-alpha.13` → `v5.0.0-alpha.13`), so dispatching with a version number checks out its release tag instead of failing at checkout
- **scripts/test-installed-version/services.js**: point `SOCKETHUB_CONFIG` at the same CI config docker-compose mounts, unblocking the feeds/metadata loopback fixtures

## Testing

- YAML validated; `biome check` clean on `services.js`; module loads and the config path resolves
- After merge, dispatching **Test NPM Package** with ref `v5.0.0-alpha.13` will still fail (the tag's harness predates the fix); the fix takes effect for master and future tags — will verify with a dispatch against master

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved release-package testing by using a consistent version tag format and validating unsupported runtime settings earlier, so failures surface clearly instead of silently skipping work.
  * Ensured the installed-version test environment starts with the correct configuration, making local package checks more reliable.

* **Chores**
  * Narrowed one release-publishing workflow to the Node runtime for more predictable test execution.

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -25,7 +25,7 @@ jobs:
       (github.event_name == 'workflow_dispatch')
     uses: ./.github/workflows/test-npm-package.yml
     with:
-      runtime: both
+      runtime: node
       suite: all
 
   publish:
--- a/.github/workflows/test-npm-package.yml
+++ b/.github/workflows/test-npm-package.yml
@@ -53,6 +53,19 @@ jobs:
       fail-fast: false
 
     steps:
+      # node is the only supported deployment runtime. An unrecognized value
+      # must fail the job immediately, not skip steps: a caller passing a
+      # stale value (e.g. 'both') would otherwise produce a green job that
+      # tested nothing.
+      - name: Validate runtime
+        env:
+          INPUT_RUNTIME: ${{ inputs.runtime || github.event.inputs.runtime || 'node' }}
+        run: |
+          if [ "$INPUT_RUNTIME" != "node" ]; then
+            echo "Error: unsupported runtime '$INPUT_RUNTIME' (supported: node)" >&2
+            exit 1
+          fi
+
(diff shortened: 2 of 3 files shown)


silverbucket:
Verified end-to-end: dispatched Test NPM Package from this branch — all 9 suites pass, including the two that failed against v5.0.0-alpha.13 (`browser-basic`, `browser-contract`): https://github.com/sockethub/sockethub/actions/runs/28738803829

PR #1170, opened 2026-07-05
chore(server): upgrade to Express 5 and drop body-parser

silverbucket:
## Summary
- Upgrade `express` 4.22 → 5.x (and `@types/express` → v5)
- Replace the standalone `body-parser` dependency with the bundled `express.json()` / `express.urlencoded()`
- Fix the one Express 5 breaking change in our code: the `app.get("*")` examples SPA fallback

## Why

Express 5 forwards rejected promises from `async` route handlers to error middleware, which removes an entire class of process-crashing unhandled rejections (Express 4 does not catch these). `body-parser` is exactly what Express bundles as `express.json()`/`express.urlencoded()`, so the separate dependency is redundant.

## Changes

- `express` → `^5.1.0`, `@types/express` → `^5.0.0`; removed `body-parser` and `@types/body-parser`.
- `listener.ts`: `bodyParser.urlencoded/json` → `express.urlencoded/json`; `app.get("*", …)` → `app.get(/.*/, …)` (path-to-regexp v8 no longer accepts the bare `"*"` string).
- Literal resource routes in `routes.ts` are unaffected (no wildcard/`:param` patterns).

## Verification

- Full unit suite passes (610 tests).
- Booted the server and confirmed base resource routes, the examples SPA fallback (the new regex route), and the socket.io endpoint all serve 200 with no startup errors.
- socket.io rate-limiter integration suite passes against the Express 5 server.

## Notes

The related `rate-limit-redis` idea (shared cross-instance rate limits) is intentionally **not** in this PR. Master's only `express-rate-limit` user is the examples static-file limiter, where per-instance in-memory limiting is actually the correct behavior — the endpoint that benefits from shared limits is the HTTP actions API on #1036, so `rate-limit-redis` will land there instead.

Once this merges, #1036 (HTTP actions) will rebase onto Express 5 and can drop its defensive `try/catch` around the async POST handler, since Express 5 handles async rejections natively.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -9,10 +9,10 @@
         "@esm-bundle/chai": "^4.3.4-fix.0",
         "@lerna-lite/cli": "4.9.0",
         "@lerna-lite/publish": "4.9.0",
-        "@sockethub/client": "workspace:5.0.0-alpha.12",
-        "@sockethub/data-layer": "workspace:1.0.0-alpha.12",
-        "@sockethub/examples": "workspace:1.0.0-alpha.12",
-        "@sockethub/schemas": "workspace:3.0.0-alpha.12",
+        "@sockethub/client": "workspace:5.0.0-alpha.13",
+        "@sockethub/data-layer": "workspace:1.0.0-alpha.13",
+        "@sockethub/examples": "workspace:1.0.0-alpha.13",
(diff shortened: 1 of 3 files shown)


silverbucket:
@coderabbitai code review

PR #1171, opened 2026-07-05
fix(data-layer): expire and tear down credential-store Redis keys

### LINKED ISSUE #1169: data-layer: credential-store Redis keys never expire or get deleted (unbounded growth)
## Summary

Credentials written to Redis via `CredentialsStore` (`secure-store-redis` `SecureStore`) have **no TTL and are never deleted**. `SecureStore.save()` issues a plain `hset` with no expiry, and no code path in the server removes credential-store keys. Over time this accumulates orphaned, encrypted credential blobs in Redis — an unbounded storage-growth / resource-exhaustion vector.

This is **live and reproducible today**. Scanning a running instance:

```
$ redis-cli --scan --pattern '*credentials-store*'
sockethub:FQGFo^B4ZECoVGTB:data-layer:credentials-store:kayuekiAqV4XLLeKAAAB
sockethub:#1X@gEVQdSi59oW$:data-layer:credentials-store:hcsEvuXdW5jzHOfcAAAB
sockethub:SJDtIhr^PG2jf8CO:data-layer:credentials-store:i-dfmJ8DWXnH4MhAAAAB
```

These are leftover socket-session credential hashes (`sessionId` = socket.io id). Because `parentId` is randomized per server boot, every restart also strands the previous run's keys forever under a now-unused `parentId` prefix.

## Scope

- **Socket sessions**: one leaked key per connection (and per reconnect, since `socket.id` changes). Bounded per run but never cleaned; orphaned across restarts.
- **HTTP actions**: *was* one leaked key per credential-bearing request (unique single-use session id). **Fixed** in PR #1036 via `CredentialsStore.teardown()` called from the request cleanup — see that PR for the pattern. This issue covers the remaining/general case.

## Not a confidentiality breach

Each blob is AES-encrypted with a per-session secret derived from `parentSecret1` (also randomized per boot) + a random per-session secret that is never persisted. Once the session ends the decryption secret is gone, so the data is unreadable. The problem is **unbounded Redis memory growth** → eviction of live keys / OOM → availability impact.

## Requested fix

Ensure credential-store keys are removed deterministically. Options, roughly in order of preference:

1. **TTL that is extended on use** (sliding expiry): set an expiry on the credential hash on every `save`/`get` (e.g. `PEXPIRE`), so active sessions stay alive and abandoned ones self-evict. This is the most robust — it survives crashes and ungraceful shutdowns where explicit cleanup never runs. `secure-store-redis` exposes the underlying client; `SecureStore` may need a small TTL option or we set it from `CredentialsStore`.
2. **Explicit teardown on session end**: delete the key when a socket disconnects (wire `CredentialsStore.teardown()` — added in PR #1036 — into the disconnect/`cleanupClient` path) and on server `shutdown()`. Necessary but not sufficient alone, since a crash skips it.
3. Ideally **both**: sliding TTL as the backstop, explicit teardown for promptness.

Also consider a one-time startup sweep of `sockethub:*:data-layer:credentials-store:*` keys whose `parentId` is not the current boot's, to reclaim already-stranded keys.

## Acceptance criteria

- Credential keys for ended sessions do not survive indefinitely (bounded by a TTL and/or removed on disconnect).
- Server `shutdown()` removes this instance's credential keys.
- Orphaned keys from prior boots are reclaimable (sweep and/or TTL).
- Regression coverage for socket disconnect cleanup and TTL refresh-on-use.

## References

- `packages/data-layer/src/credentials-store.ts` (`save`, `get`, new `teardown`)
- `secure-store-redis` `SecureStore.save` → `hset` with no expiry
- PR #1036 (HTTP actions) — introduced `CredentialsStore.teardown()` and the HTTP-side fix

silverbucket:
Closes #1169.

Builds on `CredentialsStore.teardown()` from #1036 (merged); rebased onto master.

## What

Credential blobs written through `SecureStore` accumulated in Redis forever: no TTL, no deletion on socket disconnect, and every boot randomizes `parentId`, stranding the previous run's keys permanently (see the live scan in #1169). This fixes the key lifecycle end to end, adopting the `ttl` + `deleteAll()` APIs shipped in secure-store-redis 4.2.0 (silverbucket/secure-store-redis#127) for exactly this purpose.

- **Sliding TTL** — new `credentials.ttlMs` config (default 7 days, `0` disables, env `SOCKETHUB_CREDENTIALS_TTL_MS`), applied at every `CredentialsStore` construction site. SecureStore refreshes the expiry on every save and read; the TTL is forwarded to platform child processes via env so their per-job credential reads refresh it too. Sized generously on purpose: idle-but-connected sessions don't refresh it, so it's a crash backstop, not a session timeout.
- **Teardown on disconnect** — the socket `disconnect` handler now deletes the session's credential key. A reconnect gets a new `socket.id` and re-sends credentials, so nothing ever reads the old key again.
- **Purge on graceful shutdown** — `shutdown()` SCAN+DELs keys under the current `parentId` only. No cross-parentId sweep: Redis may be shared by multiple running instances, so keys under other parentIds are never touched. The SCAN pattern glob-escapes the parentId.
- **Redis reconnect robustness** (from the follow-up comment on #1169) — shared connections previously stopped reconnecting after 3 attempts, leaving credentials/idempotency permanently broken after a brief Redis outage. They now retry forever with capped backoff; commands still fail fast while disconnected.
- **`crypto.randId()`** — strictly alphanumeric identifier generator, used for `parentId`, HTTP session ids, and server-minted request ids. `randToken()`'s charset includes Redis glob and shell metacharacters (live parentIds like `#1X@gEVQdSi59oW$`), and server-minted request ids could previously fail the endpoint's own `REQUEST_ID_PATTERN` validation. `randToken()` remains the generator for secrets.

## Behavior note

A job still queued when its client disconnects now fails its credential lookup (previously the orphaned key kept it readable). The disconnected client could never receive the result, so this is accepted.

Keys orphaned by versions prior to this fix are **not** auto-reclaimed; remove them manually with e.g. `redis-cli --scan --pattern '*credentials-store*' | xargs -L1 redis-cli del` if desired.

## Verification

Beyond unit tests (645 pass), verified at runtime against local Redis with `SOCKETHUB_CREDENTIALS_TTL_MS=120000` and scripted socket.io clients:

- key appears on credential save with `PTTL ≈ 120000` (env mapping confirmed; default is 7d)
- PTTL counts down and jumps back on credential re-save (sliding refresh on **save**)
- PTTL jumps back when an XMPP `connect` job makes the platform child read credentials (refresh on **read**, proving env forwarding through the fork boundary)
- key deleted immediately on socket disconnect
- SIGINT with two live session keys and no disconnects fired → both keys purged; pre-existing orphan keys from other parentIds untouched
- SIGKILL'd instance leaves keys carrying their TTL, which self-evict

## Summary by CodeRabbit

* **New Features**
  * Added configurable sliding expiry for session-stored credentials (`credentials.ttlMs`), with support via app settings and environment variable.
  * Added new credential purge utilities to remove stored credentials for a given parent identifier.
  * Improved generated identifiers (`randId`) to be safe for use in keys, headers, and similar contexts.

* **Bug Fixes**
  * Improved cleanup of credential-store data on shutdown and socket disconnect, using a safer best-effort deletion flow.
  * Updated Redis connection behavior to simplify retry and support offline queuing for relevant connections.


silverbucket:
@coderabbitai code review

silverbucket:
@coderabbitai code review

silverbucket:
@coderabbitai code review

Issue #1166, opened 2026-07-05
platform crash-recovery race: stale instance shutdown obliterates replacement's queue and evicts it from the instance map


silverbucket:
## Occurrences (2x on 2026-07-05, unrelated PRs)

- PR #1164 (workflow-only change): https://github.com/sockethub/sockethub/actions/runs/28739155768 — passed on re-run
- PR #1165 (workflow-only change): https://github.com/sockethub/sockethub/actions/runs/28740718493 — same signature

Both PRs touch only `.github/workflows/` and `scripts/test-installed-version/`, which the `integration:process` job never loads, and the identical job passed on adjacent runs (e.g. #1104's run minutes apart).

## Failure signature

Always the same three tests in `integration/process.integration.ts`, in order:

```
(fail) Parent Process Sudden Termination > Dummy platform crash detection > should recover from SIGTERM [~8030ms]
(fail) Parent Process Sudden Termination > Dummy platform crash detection > should recover from heartbeat timeout (hang) [0.10ms]
(fail) Parent Process Sudden Termination > should terminate child process when parent dies suddenly [0.08ms]
```

The first test hits its timeout (`config.timeouts.process + config.timeouts.cleanup + 5000` ≈ 8s): the dummy platform child either doesn't exit or doesn't restart within `config.timeouts.process` under CI load. The next two tests then fail in <0.1ms — cascade failures from the broken suite state (dummy platform never restarted), not independent failures.

## Possible directions

- Bump `config.timeouts.process` headroom for the restart wait in CI (the 200ms poll in `waitForProcessExit` plus janitor scheduling can plausibly exceed ~3s on a loaded runner)
- Make the two follow-on tests re-ensure the dummy platform (`ensureDummyPlatform()`) instead of inheriting broken state, so one timeout doesn't fail three tests

silverbucket:
Dug into attempt 1 of run [28739155768](https://github.com/sockethub/sockethub/actions/runs/28739155768) (job `process`, id 85218700217) and this is **not a timing/CI-load flake** — the logs plus server code point to a real crash-recovery race in the server. Reframing below.

## Corrections to the original diagnosis

[...]

PR #1175, opened 2026-07-06
fix(server): serialize platform replacement behind crashed instance teardown

silverbucket:
## Problem

The `process` integration job flaked on the release PR (#1173, [run 28787974839](https://github.com/sockethub/sockethub/actions/runs/28787974839/job/85360038426)): `should recover from uncaught TypeError` timed out waiting for the platform child to exit, and the three tests after it cascade-failed.

Root cause is a race in crash teardown that #1166/#1167 narrowed but didn't close:

1. A platform child crashes; its `close` handler starts `shutdown()`, which checks for a replacement **once, at teardown start**, finds none, and proceeds to `queue.shutdown()` — BullMQ `pause()` + `obliterate({force: true})` on a Redis queue keyed by identifier and therefore shared with any future replacement.
2. While that pause/obliterate is still in flight, a message for the same identifier arrives. `ensureProcess` creates the replacement and its queue under the same Redis name. `markReplaced()` fires too late — the in-flight teardown already passed its guard.
3. Jobs enqueued for the replacement are destroyed by the old teardown. In the flaky run, the exit1 test's restart poll created the replacement 9ms after the close event, and the next test's `throwTypeError` job silently vanished — the child never crashed, so `waitForProcessExit` timed out.

## Fix

- **`PlatformInstance.shutdown()` is single-flight**: every caller shares one teardown run. Awaiting `shutdown()` now means the pause/obliterate has actually finished, rather than a second, independent no-op run having resolved alongside the first. (This also collapses the previously duplicated teardown from `handleProcessClose` → `broadcastFatalError` → `shutdown()` → `shutdown()`.)
- **`ensureProcess` awaits the dead predecessor's teardown** before creating the replacement, so an in-flight obliterate can never land on the replacement's queue.
- Because the path now yields the event loop, **runs are chained per identifier** — two concurrent messages for the same dead instance can no longer each fork a replacement child.
- `ProcessManager.get()` becomes async; its only call site (the shared socket/HTTP message middleware) is already an async function.

## Testing

- New unit test pins the ordering: replacement creation blocks until an in-flight teardown resolves.
- New unit test covers concurrent requests for a dead instance yielding a single replacement.
- New unit test covers single-flight `shutdown()`.
- All 175 server unit tests pass; biome clean.
- Runtime-verified against a locally launched server + Redis: echo → `exit1` crash → recovery → `throwTypeError` crash → recovery, all behave correctly.
- The `process` integration suite doesn't run on macOS (pre-existing `ps` discovery incompatibility, fails identically on master), so CI is the gate for the integration path.

## Summary by CodeRabbit

* **New Features**
  * Improved handling of platform process reuse and replacement, making concurrent requests more reliable.
  * Shutdown operations are now single-flight, so repeated shutdown calls share the same in-progress teardown.

* **Bug Fixes**
  * Fixed race conditions that could start multiple platform processes at once.
  * Improved replacement timing to avoid issues when a dead process is being torn down.
  * Updated message handling to properly wait for platform lookup before continuing.

### DIFF
--- a/packages/server/src/message-handlers.ts
+++ b/packages/server/src/message-handlers.ts
@@ -147,9 +147,11 @@ export function createMessageHandlers(
                     );
                     return;
                 }
-                let platformInstance: ReturnType<ProcessManager["get"]>;
+                let platformInstance: Awaited<
+                    ReturnType<ProcessManager["get"]>
+                >;
                 try {
-                    platformInstance = processManager.get(
+                    platformInstance = await processManager.get(
                         platformId,
                         msg.actor.id,
                         platformSessionId,
--- a/packages/server/src/platform-instance.test.ts
+++ b/packages/server/src/platform-instance.test.ts
@@ -166,6 +166,21 @@ describe("PlatformInstance", () => {
             expect(platformInstances.has("platform identifier")).toBeFalse();
         });
 
+        test("shutdown is single-flight: repeated calls share one teardown", async () => {
+            pi.initQueue("a secret");
+            const queue = pi.queue;
+            const first = pi.shutdown();
+            const second = pi.shutdown();
+            // both callers must observe the same teardown run; a second,
+            // independent run would resolve while the first is still
+            // obliterating the shared Redis queue
+            expect(second).toBe(first);
+            await first;
+            sandbox.assert.calledOnce(queue.shutdown);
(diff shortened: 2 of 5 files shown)


silverbucket:
@coderabbitai code review

PR #1179, opened 2026-07-06
fix(server): share and harden multi-origin CORS resolution

### LINKED ISSUE #1176: Document ability to support multiple CORS origins
CORS Origin by design is only for one domain `example.com`, or all with `*`. Sockethub allows for multiple domains to be specified, so it can be a service for one or more trusted webapps.

However it's not clearly documented at this time

silverbucket:
## Summary

Multi-origin CORS support (the ask in #1176) already landed for socket.io in #1158 and for the HTTP actions endpoint in #1036 — `sockethub:cors:origin` accepts a comma-separated list and the matching request origin is echoed back dynamically. This PR closes the remaining gaps:

- **One shared parser.** `listener.ts` and `http/actions.ts` each parsed the origin list independently; both now resolve through a single `parseCorsOrigins()` (`packages/server/src/cors.ts`) so the two transports can never drift apart.
- **Origin normalization.** Matching was exact string comparison, but browsers send the `Origin` header lowercase with no trailing slash and default ports stripped — so a configured `https://App.example.com/` silently never matched. Configured entries are now normalized to the browser serialization, invalid entries (no scheme, opaque origin) are dropped with a logged warning, and a `*` inside a list is treated as allow-any with a warning.
- **Fail closed.** A restrictive config that yields no valid origins now blocks all cross-origin requests instead of silently falling back to `*`.
- **Env/CLI bindings.** `sockethub:cors:origin` was config-file-only; it can now be set via `SOCKETHUB_CORS_ORIGIN` or `--cors.origin` — the natural fit for the multi-tenant service deployment described in #1176.
- **Docs.** `docs/configuration.md` now covers the setting under Server Settings, the CORS section, and Environment Variables.

Deliberately **not** included: wildcard-subdomain/regex patterns (exact-match lists are the safest primitive and the issue doesn't ask for them), and the `Access-Control-Allow-Origin: *` on the static client-JS routes (`routes.ts`), which is correct for public credential-free assets.

## Runtime verification

Started the built server with `SOCKETHUB_CORS_ORIGIN="https://App.Example.com/, http://localhost:3000"`:

- polling handshake with `Origin: http://localhost:3000` → echoed in `Access-Control-Allow-Origin` + `Vary: Origin`; full socket.io client connect succeeded
- `Origin: https://app.example.com` → allowed despite the uppercase/trailing-slash config entry (normalization warning logged at boot)
- `Origin: https://evil.example` → no `Access-Control-Allow-Origin` header emitted (browser blocks)

## Tests

`bun run lint`, `bun run build`, `bun test` (660 pass) — new unit coverage in `cors.test.ts` plus normalization/fail-closed cases in `actions.test.ts`.

Closes #1176

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -76,10 +76,21 @@ Sockethub uses a JSON configuration file:
   "sockethub": {
     "port": 10550,          // Port Sockethub listens on
     "host": "localhost",    // Bind address
-    "path": "/sockethub"    // WebSocket endpoint path
+    "path": "/sockethub",   // WebSocket endpoint path
+    "cors": {
+      "origin": "https://app.example.com, https://admin.example.com"
+    }
   }
 }
 ```
+
+`sockethub:cors:origin` (default `"*"`) controls which web origins may connect
+— it governs both socket.io connections and the HTTP actions endpoint. It
+accepts `"*"`, a single origin, or a comma-separated list of origins; when a
+list is configured, the matching request origin is echoed back in
+`Access-Control-Allow-Origin`. See [CORS](#cors) for details. It can also be
+set with the `SOCKETHUB_CORS_ORIGIN` environment variable or the
+`--cors.origin` command-line flag.
 
 ### Public Settings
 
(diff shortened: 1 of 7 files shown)


silverbucket:
@coderabbitai code review

silverbucket:
Both review points addressed in 7d10366f:

- **Per-request re-parse/log spam**: `createCorsMiddleware()` now calls `parseCorsOrigins()` once at route registration and closes over the resolved `"*" | string[]`; the per-request handler only does the match. Runtime-verified with the built server: the normalization warning now appears exactly twice in the log (socket.io listener parse + actions middleware parse at boot) regardless of request count.
- **Request-origin normalization**: `resolveAllowedOrigin()` moved to `cors.ts`, takes the pre-parsed allow-list, and runs the request `Origin` header through the same `new URL(...).origin` normalization as configured entries (malformed/opaque origins fail closed) instead of assuming clients send the canonical serialization.

Tests moved/extended accordingly (`cors.test.ts` now covers both functions plus the malformed-origin and empty-allow-list paths); full suite passes (662).

silverbucket:
@coderabbitai code review

PR #1203, opened 2026-08-10
fix(examples): show batch response log details

### LINKED ISSUE #1201: CalDAV example: event payloads for fetched calendars are empty/non-existent
Fetch calendars, click "Details" next to the events in the log, the modal content is blank.

silverbucket:
## Summary

- retain the originating request ID in activity-log metadata for batch responses
- resolve modal content through that explicit relationship instead of parsing synthetic IDs
- add regression coverage for a CalDAV-style URL item ID containing hyphens

## Root cause

Collection entries use synthetic log keys derived from response item IDs. The details handler attempted to recover the parent request by splitting those keys on hyphens. URL-shaped CalDAV IDs therefore resolved to a nonexistent log entry and threw after opening the modal, leaving it blank.

Fixes #1201.

## Validation

- `bun run test:unit` in `packages/examples`
- `bun run check` in `packages/examples`
- `bun run build` in `packages/examples`
- `bun run lint`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved log detail matching so responses are correctly associated with their originating requests, including calendar batch responses.
  * Enhanced log views to reliably display corresponding sent requests and responses.

* **Tests**
  * Added coverage verifying request and response association for URL-based calendar batch responses.

### DIFF
--- a/packages/examples/src/components/logs/Logger.svelte
+++ b/packages/examples/src/components/logs/Logger.svelte
@@ -1,6 +1,7 @@
 <script lang="ts" module>
 import type { AnyActivityStream } from "$lib/sockethub";
 import { writable } from "svelte/store";
+import { resolveLogDetails, type LogMetadata } from "./log-details";
 
 type LogEntries = Record<
     string,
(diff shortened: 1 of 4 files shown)


raucao (approved):
Confirming that it's fixed for CalDAV here

PR #1210, opened 2026-08-13
fix(server): enable resource caps by default

silverbucket:
## Summary

- limit each client IP to 20 concurrent Socket.IO connections by default
- limit the server to 100 concurrent platform child processes by default
- document how operators can override or disable the caps

## Root cause

Both resource controls existed but defaulted to disabled, allowing connection fan-out to bypass per-socket event limiting and permitting unbounded platform process creation.

## Impact

Default installations now retain bounded concurrency while operators with external controls can explicitly set either cap to zero.

## Validation

- targeted server tests (45 passed)
- `git diff --check`
- `bun run lint`

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -53,7 +53,11 @@ does not start).
   "rateLimiter": {
     "windowMs": 1000,
     "maxRequests": 100,
-    "blockDurationMs": 5000
+    "blockDurationMs": 5000,
+    "maxConnectionsPerIp": 20
+  },
+  "limits": {
+    "maxPlatformInstances": 100
   },
   "credentialCheck": {
     "reconnectIpSource": "socket",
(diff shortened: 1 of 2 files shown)


silverbucket:
@coderabbitai code review

raucao:
> limit the server to 100 concurrent platform child processes by default

Does this mean if i have 20 users with 5 platforms each, the limit is reached? :thinking:

silverbucket:
> > limit the server to 100 concurrent platform child processes by default
> 
> Does this mean if i have 20 users with 5 platforms each, the limit is reached? 🤔

Well, theoretically yes, however there are just 2 persistent Sockethub platforms. So, a user would need 5 different XMPP/IRC accounts logged in all together to reach 5.

And then there are the shared platform processes, which would only ever be one process per platform (for now). 

So, as an example, it could be something like 4 static processes (CaldDAV, CarDAV, Metadata, Feeds), 96/2 = 48 users running one IRC and one XMPP account.

I've tried to document it better here https://github.com/sockethub/sockethub/pull/1213

raucao:
So it's a development default? 48 users can't be a production default, right?

silverbucket:
> So it's a development default? 48 users can't be a production default, right?

Well anything can be changed in the sockethub config -- do you have a suggestion for what the defaults should be?

raucao:
Actually, just leave it to the docs. The default should probably be good for self-hosting.

PR #1211, opened 2026-08-13
fix(examples): escape feed entry titles

silverbucket:
## Summary

Render feed entry titles through normal Svelte text interpolation instead of `{@html}`.

## Root cause

The title used raw HTML whenever the entry content type was HTML, but that content type is derived from the feed description rather than from the title. An attacker-controlled feed could therefore supply active HTML in its title.

## Impact

Feed titles are always escaped and cannot execute markup in the examples application.

## Validation

- Svelte autofixer
- `git diff --check`
- `bun run lint`
- `bun run build`


silverbucket:
@codereviewai code review

Issue #1054, opened 2026-03-30
Auth Phase 1: Pre-shared token authentication (MVP)


silverbucket:
Parent: #735

## Overview

Implement the foundation of sockethub's token auth system using pre-shared tokens. This phase delivers the core infrastructure that all future auth backends (LDAP, remoteStorage) will build on: the token store, the Socket.IO handshake middleware, and the CLI tool.

No HTTP token endpoint in this phase — pre-shared tokens are passed directly in the Socket.IO handshake. The `POST /auth/token` endpoint is introduced in Phase 2 when LDAP/RS backends need a credential-to-token exchange.

When `auth.required` is `false` (default), behavior is identical to today — fully backward compatible.

---

### 1. Configuration

**`packages/server/src/defaults.json`** — add `auth` section:

```json
{
  "auth": {
    "required": false,
    "backends": {
      "preshared": {
        "enabled": false,
        "tokenFile": ""
      }
    },
    "rateLimiter": {
      "windowMs": 60000,
      "maxAttempts": 10,
      "blockDurationMs": 300000
    }
  }
}
```

**`packages/server/src/config.ts`** — add env var overrides:
- `AUTH_REQUIRED` → `auth:required`
- `SOCKETHUB_AUTH_TOKEN_FILE` → `auth:backends:preshared:tokenFile`

Startup validation: if `auth.required` is `true`, at least one backend must be `enabled` with valid config (for preshared: non-empty `tokenFile` pointing to an existing file). Exit with clear error message if not.

### 2. Token Store (`packages/server/src/auth/token-store.ts`)

Interface that all backends share for token validation:

```typescript
interface TokenRecord {
  hash: string;           // SHA-256 hex of the raw token
  identity: string;       // human-readable label ("webapp-prod", "bot-service")
  grantType: string;      // "preshared" | "ldap" | "rs_bearer"
  platforms?: string[];   // optional platform scoping
  createdAt: number;      // epoch ms
  expiresAt?: number;     // epoch ms, undefined = no expiry
}

interface TokenStore {
  initialize(): Promise<void>;
  lookup(tokenHash: string): Promise<TokenRecord | undefined>;
  save(record: TokenRecord): Promise<void>;
  remove(tokenHash: string): Promise<void>;
  shutdown(): void;
}
```

**`FileTokenStore`** implementation for pre-shared tokens:
- Reads JSON file at `auth.backends.preshared.tokenFile` on `initialize()`
[...]

PR #1222, opened 2026-08-16
fix(server): verify credentials match before attaching to a shared platform instance

silverbucket:
## Summary

Persistent platform instances (IRC, XMPP) are shared server-wide, keyed by `hash(platform + actor.id)`. The guard on that sharing only checked that the attaching session's credentials contained a **non-empty** password or token — it never compared them against the credentials that opened the connection.

Since each session stores its own credentials, the attaching client supplies the value being checked. **Knowing a connected user's `actor.id` plus any non-empty password was enough to join their live connection** — sending as them, and receiving their inbound traffic (private messages included), since platform emits fan out to every registered session.

`join`, `send`, `leave` and `query` are not in IRC's `requireCredentials`, so no later check caught it either.

## Verification

Before, against a real server driven by two socket.io clients (session B on a different IP with a different password):

```
added irc-0 connect to queue          queue:…:0b033cd6…   (session A)
get credentials for victim@…          credentials-store:…:rN6u…AAAD   (session B)
added irc-1 join to queue             queue:…:0b033cd6…   (session B → A's instance)
```

Same three-case harness against the real `CredentialsStore` and real middleware, before and after:

| Second session stores | Before | After |
|---|---|---|
| nothing | BLOCKED — `username already in use` | BLOCKED — `username already in use` |
| `password: "totally-wrong"` | **ATTACHED** | BLOCKED — `invalid credentials for …` |
| `password: <correct>` | ATTACHED | ATTACHED |

## What changed

The exact-match comparison already existed in `CredentialsStore.get()` — it just ran in the wrong place. Only the child process passed a `credentialsHash`, only for credentialed job types, and only *after* the parent had already registered the session. The parent had no access to the child's credential state.

- **`platform.ts`** — the child publishes its `credentialsHash` to the parent over the existing IPC channel once a credentialed call succeeds (so it reflects the credentials that actually opened the connection).
- **`platform-instance.ts`** — `PlatformInstance` holds it; new `credentialsHash` control message.
- **`credential-check.ts`** — passes it to the data layer, so an attach requires an exact match. This runs **before** `registerSession()`, so a rejected attach is never registered.
- **`credentials-store.ts`** — reorders the two checks so a passwordless attach still reports "not shareable" rather than "mismatch", keeping the same-IP anonymous reconnect path reachable.

## Behaviour notes

- **Legitimate sharing is preserved.** Two clients presenting the same credentials still share one connection (covered by `irc-nickclash.integration.js`).
- **Unknown-hash window.** Before any credentialed call completes (e.g. two clients connecting concurrently), the hash is unknown and only the original non-empty rule applies. No remote connection exists at that point, so there is no traffic to expose; the first successful connect publishes the hash and later attaches are checked in full.
- **Anonymous same-IP reconnect** is unaffected.

## Tests

- `credentials-store.test.ts` — the three cases at the decision layer: no password, wrong password, correct password.
- `credential-check.test.ts` — the incumbent hash reaches the store; a mismatch is rejected; a mismatch is **not** eligible for the same-IP reconnect escape hatch.
- `platform-instance.test.ts` — the `credentialsHash` IPC message sets the field and is not forwarded to clients.

`bun test` — 823 pass, 0 fail. `biome check` clean. `bun run build` clean.

## Follow-ups (not in this PR)

1. **Detach on failure** — a session rejected by the child's credential check stays in `platformInstance.sessions` until the janitor notices its socket is gone (`janitor.ts:62` is the only removal path). Narrower now that attach is gated, but still worth closing.
2. **Presence oracle** — `"username already in use"` is only reachable when an actor already has a live connection, so it discloses who is connected. Flattening the three failure modes is a separate change.
3. **Cross-platform credential collisions** — credentials are keyed by `actor.id` alone, with no platform component, both server-side (`store-credentials.ts:15`) and in the client's replay maps (`sockethub-client.ts:531,553`).

Context: came out of reviewing #1131. The defect predates that PR and is independent of the ID format debate.

## Summary by CodeRabbit

- **Bug Fixes**
  - Persistent shared sessions now require matching credentials when attaching or reconnecting.
  - Missing credentials are clearly rejected when a session cannot be shared without them.
  - Incorrect credentials consistently produce a mismatch error, including same-network reconnects.
  - Credential validation is now reliably synchronized across persistent platform instances.

### DIFF
--- a/packages/data-layer/integration/redis.integration.ts
+++ b/packages/data-layer/integration/redis.integration.ts
@@ -39,7 +39,9 @@ const creds: CredentialsObject = {
         password: "secret",
     },
 };
-const credsHash = "267a747d006c9a2c2e94b2f7d646400ba16e5709";
+// SHA-256 of `creds.object` (see Crypto.objectHash).
+const credsHash =
+    "8ab6f374809a069edf089497d3d091c467699fac4f6330dd056f662887995891";
 const testSecret = "aB3#xK9mP2qR7wZ4cT8nY6vH1jL5fD0s";
 
 describe("CredentialsStore", () => {
--- a/packages/data-layer/src/credentials-store.test.ts
+++ b/packages/data-layer/src/credentials-store.test.ts
@@ -209,6 +209,67 @@ describe("CredentialsStore", () => {
             });
         });
 
+        // Regression: a session attaching to an existing persistent instance
+        // must prove it holds the *same* credentials that opened it. Before
+        // this, any non-empty password was accepted, so knowing an actor id
+        // was enough to join someone else's live connection.
+        describe("session-share attach against an incumbent hash", () => {
+            const INCUMBENT_HASH = "hash-of-the-connected-credentials";
+
+            it("rejects an attach with no password at all", async () => {
+                MockObjectHash.returns(INCUMBENT_HASH);
+                MockStoreGet.returns({
+                    object: { type: "credentials", nick: "victim" },
+                });
+                try {
+                    await credentialsStore.get("an actor", INCUMBENT_HASH, {
+                        validateSessionShare: true,
+                    });
+                    expect(false).toEqual(true);
+                } catch (err) {
+                    // Reported as "not shareable" rather than "mismatch" so the
+                    // same-IP anonymous reconnect path stays reachable.
+                    expect(err).toBeInstanceOf(CredentialsNotShareableError);
+                    expect(err.toString()).toEqual(
+                        "Error: username already in use",
+                    );
+                }
+            });
+
+            it("rejects an attach with a wrong password", async () => {
+                MockObjectHash.returns("hash-of-the-attackers-credentials");
+                MockStoreGet.returns({
+                    object: { type: "credentials", password: "totally-wrong" },
+                });
+                try {
+                    await credentialsStore.get("an actor", INCUMBENT_HASH, {
+                        validateSessionShare: true,
+                    });
+                    expect(false).toEqual(true);
+                } catch (err) {
+                    expect(err).toBeInstanceOf(CredentialsMismatchError);
+                    expect(err.toString()).toEqual(
+                        "Error: invalid credentials for an actor",
+                    );
+                }
+            });
+
+            it("allows an attach with the correct password", async () => {
+                MockObjectHash.returns(INCUMBENT_HASH);
+                MockStoreGet.returns({
+                    object: { type: "credentials", password: "correct-horse" },
+                });
+                const res = await credentialsStore.get(
+                    "an actor",
+                    INCUMBENT_HASH,
+                    { validateSessionShare: true },
+                );
+                expect(res).toEqual({
+                    object: { type: "credentials", password: "correct-horse" },
+                });
+            });
+        });
+
         it("rejects credentials without password or token for session-share validation", async () => {
             MockStoreGet.returns({
                 object: { type: "credentials" },
--- a/packages/data-layer/src/credentials-store.ts
+++ b/packages/data-layer/src/credentials-store.ts
@@ -290,16 +290,10 @@ export class CredentialsStore implements CredentialsStoreInterface {
             );
         }
 
-        if (credentialsHash) {
-            // If a hash is provided, credentials must match exactly. This blocks
-            // "same actor, different credentials" reuse attempts.
-            if (credentialsHash !== this.objectHash(credentials.object)) {
-                throw new CredentialsMismatchError(
-                    `invalid credentials for ${actor}`,
-                );
-            }
-        }
-
+        // Ordered before the hash comparison so a passwordless attach always
+        // reports "not shareable" rather than "mismatch", whatever the
+        // incumbent's credentials were. Callers distinguish the two: only the
+        // not-shareable case is eligible for same-IP anonymous reconnect.
         if (options.validateSessionShare) {
             const password = credentials.object.password;
             const token = credentials.object.token;
@@ -312,6 +306,16 @@ export class CredentialsStore implements CredentialsStoreInterface {
             if (!hasPassword && !hasToken) {
                 throw new CredentialsNotShareableError(
                     "username already in use",
+                );
+            }
+        }
+
+        if (credentialsHash) {
+            // If a hash is provided, credentials must match exactly. This blocks
+            // "same actor, different credentials" reuse attempts.
+            if (credentialsHash !== this.objectHash(credentials.object)) {
+                throw new CredentialsMismatchError(
+                    `invalid credentials for ${actor}`,
                 );
             }
         }
--- a/packages/server/src/middleware/credential-check.test.ts
+++ b/packages/server/src/middleware/credential-check.test.ts
@@ -1,4 +1,12 @@
-import { afterEach, beforeAll, beforeEach, describe, expect, test } from "bun:test";
+import {
+    afterEach,
+    beforeAll,
+    beforeEach,
+    describe,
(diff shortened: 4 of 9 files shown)


silverbucket:
Assessed both reviews. One finding is valid and is now fixed; one is incorrect.

## ✅ Valid — attach failed open during the handshake window (High)

@cursor and @coderabbitai both flagged this, and both are right. My comment claiming "there is no traffic to expose" was wrong. I traced the path again:

1. `credentialsHash` is undefined until the child's first credentialed call succeeds — that's the whole remote handshake.
2. During it, `hasOtherSession` is true but the hash is unset, so only the non-empty-secret rule ran and the attach was allowed.
3. **Registration is sticky.** `janitor.ts:62` is the only other `sessions.delete` call site.
4. **An already-registered session skips this middleware entirely** on later messages, because `hasOtherSession` requires `!existing.sessions.has(socketId)`.

So a session that slipped in mid-handshake stayed attached and started receiving traffic the moment the hash arrived. Point 4 is the part I missed — I'd assumed later messages would be re-checked.

Fixed in `eb68ef23`: for persistent platforms the middleware now waits (bounded, 10s) for the incumbent's hash and compares, refusing the attach if it never arrives.

On the two suggested remedies:

- **"Fail closed when the hash is unknown"** — plain rejection breaks a legitimate case: two clients connecting concurrently with the same credentials, which `irc-nickclash.integration.js` asserts must both succeed. Waiting preserves that (the second waits for the first) while still failing closed when nothing arrives.
- **"Delay `registerSession()` until the hash is published"** — same effect, but the wait lives in the middleware where the decision already is, rather than restructuring `ProcessManager`.

Scoped to `config.persist`: non-persistent platforms never publish a hash and have no long-lived connection to join, so they keep the original rule and never wait. Without that scoping every second session on `feeds`/`metadata` would have blocked for the full timeout. Waiters are also released on teardown so a dying connection fails an attach immediately.

New tests: hash-never-arrives refuses; hash-arrives-and-matches admits; non-persistent doesn't wait; plus `waitForCredentialsHash` resolve/timeout/already-known.

## ❌ Not valid — "the published verifier is only seven hexadecimal characters"

@coderabbitai — this conflates two different functions. `Crypto.hash()` truncates to 7 chars, but the published verifier uses `Crypto.objectHash()`, which delegates to the `object-hash` package's default (full SHA-1). Verified:

```
objectHash = 90c59e240b0a86cda1419dd4607834183ccfd3f9   length = 40
hash()     = 8b45e4b                                    length = 7
```

40 hex chars, not 7. No collision concern, and no change made.

## ✅ Nit — `test()` → `it()`

Correct per `AGENTS.md`. Converted the tests this branch adds. Left the ~40 pre-existing `test()` calls in `platform-instance.test.ts` alone — a file-wide sweep is unrelated churn for this PR.

---

`bun test` 829 pass / 0 fail, `biome check` clean, build clean.

silverbucket:
Thanks — thorough review. One correction on the blocking item, then the rest are addressed.

## The blocking finding was already fixed

This review describes `615dbb62`, including the "no connection exists yet, so there is no traffic to expose" rationale. That comment and the fail-open behaviour it justified were removed in **`eb68ef23`**, pushed before this review. The analysis was right — steps 1–7 are exactly the path, and step 6 ("never rechecked or removed") is the part I originally missed, since `hasOtherSession` requires `!sessions.has(socketId)` so an attached session skips the middleware on later messages.

Current behaviour (`credential-check.ts`): for `config.persist` platforms, an attach with an unknown incumbent hash **waits** for it (bounded, 10s) and then compares, refusing if it never arrives.

On the specific recommendations 1–4:

1. **"Treat unknown hash as pending, not shareable"** — done.
2. **"Defer the second attachment until authentication finishes"** — done, via `PlatformInstance.waitForCredentialsHash()`.
3. **"Register only after successful comparison"** — already holds: `credentialCheck` runs before `processManager.get()`, so a refused attach is never registered.
4. **"Require acknowledgment the parent stored the hash before reporting connect complete"** — not needed, and worth being precise about why. The IPC-lag race you describe is real in `615dbb62`, but the wait closes it rather than papering over it: a session arriving in that gap doesn't read a stale `undefined` and proceed, it blocks on the waiter, which the parent resolves when it processes the message. An ack handshake would add a round-trip to every connect to close a window that is already closed from the reader's side.

I chose waiting over plain rejection because rejection breaks a legitimate case — two clients connecting concurrently with the same credentials, which `irc-nickclash.integration.js` asserts must both succeed. Scoped to `persist` because non-persistent platforms never publish a hash and would otherwise block for the full timeout.

## 5. Hostile concurrent-connect test — added

Fair: I had "hash never arrives" and "hash arrives and matches", but not the hostile variant. Added in `899b7512` — a session racing the incumbent's connect with a *different* password waits for the hash and is then refused. Also covered: non-persistent platforms don't wait, and `waitForCredentialsHash` resolve/timeout/already-known.

## 6. Exact vs normalized credentials — documented, not changed

Good catch, and you're right that `secure` omitted vs `secure: true` will not share. But this is not introduced here: the child has always compared `platform.credentialsHash` on `connect`/`update` (`platform.ts:394-399` → `credentials-store.ts:312-320`), so a second session with equivalent-but-different credentials already failed to connect. What this PR changes is that it can no longer `join`/`send` either — which was the vulnerability.

So the contract is pre-existing; it was just undocumented and inconsistently applied. Now stated explicitly in the middleware docblock. Normalizing to effective connection settings would mean deciding, per platform, which fields are semantically equivalent — a widening of an authorization check that deserves its own change, not a rider on a security fix.

## 7. SHA-1 → SHA-256 — changed

Correct, `object-hash` defaults to SHA-1. Pinned `Crypto.objectHash()` to SHA-256 in `899b7512`, and updated the one hard-coded fixture in the redis integration suite.

Agreed on the severity framing: the value is never exposed to clients and forging a match needs a second preimage of an unknown credential object, which SHA-1 doesn't give you. Hygiene, not an exploit fix — but not worth arguing about for a one-line change.

## Anonymous same-IP reconnect

Agreed on all counts, including that it's not introduced here. Client IP is weak identity, and behind NAT it's close to none. Now called out in the code with a pointer to #1054 — a real client identity is the thing that actually fixes it, and changing reconnect behaviour inside a security fix seemed like the wrong trade.

---

`bun test` 830 pass / 0 fail; data-layer redis integration 19 pass / 3 skip / 0 fail against local Redis; `biome check` and build clean.

raucao:
> Persistent platform instances (IRC, XMPP) are shared server-wide, keyed by `hash(platform + actor.id)`

Doesn't that prevent us from using two different clients with two different client names, e.g. mobile and desktop, with the same Sockethub instance? For IRC (pre v3 at least) it doesn't matter much, but for XMPP you want different clients to use all the per-client affordances of the XMPP protocol (e.g. read a message on one client but not another, etc.).

silverbucket:
> > Persistent platform instances (IRC, XMPP) are shared server-wide, keyed by `hash(platform + actor.id)`
> 
> Doesn't that prevent us from using two different clients with two different client names, e.g. mobile and desktop, with the same Sockethub instance? For IRC (pre v3 at least) it doesn't matter much, but for XMPP you want different clients to use all the per-client affordances of the XMPP protocol (e.g. read a message on one client but not another, etc.).

This should all be handled by the resource, so you'd have a different actor.id for each independent session (e.g. `xmpp:user@host/Phone`vs `xmpp:user@host/Laptop`). Those would be two separate platforms + socket connections. You'd only share the connection if you used the exact same resource as well.

raucao:
Ah right, thanks for the clarification.

PR #1223, opened 2026-08-16
fix(server): detach unauthorized sessions, unify share errors, scope credentials by platform

silverbucket:
Three related session/credential hygiene issues found while reviewing #1131. Each is independent of the ID-format discussion there, and independent of #1222 (branched from `master`, not stacked).

## 1. Detach on credential failure

When the child rejected a credentialed job on an already-initialized platform, only that job failed. The session stayed in `platformInstance.sessions` — and kept receiving the connection's traffic through the fan-out in `handleProcessMessage` and `broadcastToSharedPeers` — until the janitor noticed its socket had gone. `janitor.ts:62` held the only `sessions.delete` call site.

The child now emits a `sessionUnauthorized` control message naming the session, and `PlatformInstance.deregisterSession()` drops it.

Signalling explicitly, rather than inferring from job failure, is deliberate: a `connect` can also fail because the remote server is unreachable, and detaching a healthy session on a network blip would silently stop its messages. Only credential-store rejections trigger it. A session that later presents valid credentials re-registers through `ProcessManager` on its next message.

## 2. Uniform session-share rejections

`credentials not found for X`, `invalid credentials for X` and `username already in use` were three distinguishable answers to an attach attempt, so a caller probing guessed actor ids learned *why* it was refused. All attach failures now return a single `SESSION_SHARE_DENIED` message.

Error **classes** are unchanged, so behaviour that depends on them is preserved:
- same-IP anonymous reconnect still keys off `CredentialsNotShareableError`
- a missing-credentials failure is deliberately still a plain `Error`, so it stays *ineligible* for that reconnect path

Detailed messages are kept for the platform's own credential reads, where the caller already owns the credentials being described.

**This narrows but does not close presence disclosure.** Accept-vs-reject is inherently observable — an attacker can still infer that *some* session holds a given actor id. Only the reason is now hidden. Fully closing it would mean rejecting unknown actors too, which isn't viable.

## 3. Platform-scoped credential keys

Credentials were keyed by `actor.id` alone with no platform component, so one session holding credentials for the same visible id on two platforms (an `alice@example.org` that is both an IRC nick and an XMPP JID) had the second save silently overwrite the first.

New `buildCredentialsKey(platform, actor)` is used by `store-credentials`, `credential-check`, and the child's per-job read. The parent resolves the platform from `@context`, the child from the platform name it was forked with — verified to be the same string (`PlatformInstance` forks with `params.platform`, which is `resolvePlatformId(msg)`).

The client had the same bug in its reconnect replay maps (`events.credentials`, `connect`, `join`), so `SockethubClient.getKey()` is scoped by platform too.

## Tests

- `queue-id.test.ts` — the same actor id on two platforms yields distinct keys
- `credentials-store.test.ts` — all four attach-rejection paths return the identical message; detailed messages survive for non-share reads
- `platform-instance.test.ts` — `sessionUnauthorized` deregisters; a deregistered session stops receiving platform messages; unknown session is a no-op
- `store-credentials.test.ts` / `credential-check.test.ts` — updated for the platform-scoped key
- `sockethub-client.test.ts` — credentials for one actor on two platforms no longer collide

`bun test` — 826 pass, 0 fail. `biome check` clean, `markdownlint` clean on changed docs, `bun run build` clean.

## Notes for review

- **Behaviour change:** the client-visible error for a refused share is now `unable to attach session for this actor`. `docs/client-guide.md`, `docs/configuration.md` and `packages/server/README.md` are updated.
- **Credential key shape changed.** Keys are per-session and TTL'd, so nothing needs migrating, but a server restart mid-session invalidates in-flight credentials (they are re-sent on reconnect).
- **Overlaps #1222** in `credentials-store.ts` and `credential-check.ts`. Whichever lands second will need a small rebase; the changes are complementary (that PR adds the exact-match comparison, this one changes what the rejection *says* and how the key is built).

## Summary by CodeRabbit

* **Bug Fixes**
  * Credentials are now stored separately across platforms, preventing accounts with matching identifiers from overwriting one another.
  * Unauthorized sessions are automatically disconnected and prevented from receiving further platform messages.
  * Session-sharing failures now use a consistent, privacy-preserving error message without revealing the specific cause.

* **Documentation**
  * Updated client, server, and configuration guidance to reflect credential-sharing requirements and standardized rejection messages.

### DIFF
--- a/docs/client-guide.md
+++ b/docs/client-guide.md
@@ -254,7 +254,7 @@ The rejected request returns an ActivityStream response with an `error`:
   "type": "connect",
   "actor": { "id": "mynick@irc.libera.chat", "type": "person", "name": "mynick" },
   "id": "1",
-  "error": "username already in use"
+  "error": "unable to attach session for this actor"
 }
 ```
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -239,8 +239,9 @@ When another socket is already attached to that actor:
 
 - Session-share validation runs in the data layer.
 - Credentials with a non-empty `password` are considered shareable.
-- Credentials without a non-empty `password` are not shareable and return
-  `username already in use`.
+- Credentials without a non-empty `password` are not shareable.
+- Every rejection returns the same message, `unable to attach session for this
+  actor`, so the reason is not disclosed to the caller.
 
 This prevents anonymous/username-only accounts from accidentally sharing the
 same platform thread across different users.
--- a/packages/client/src/sockethub-client.test.ts
+++ b/packages/client/src/sockethub-client.test.ts
@@ -708,6 +708,33 @@ describe("SockethubClient", () => {
         });
     });
 
+    describe("replay keys are scoped by platform", () => {
+        beforeEach(() => {
+            sc.socket.connected = true;
+            sc._socket.connected = true;
+            socket.emit("schemas", TEST_REGISTRY);
+            sandbox.stub(sc, "validateActivity").returns("");
+        });
+
+        it("keeps credentials for one actor id on two platforms separate", () => {
+            // Regression: keyed by actor alone, the second platform's
+            // credentials replaced the first, so a reconnect replayed only one.
+            const actor = { id: "alice@example.org", type: "person" };
+            sc.socket.emit("credentials", {
+                "@context": sc.contextFor("test-xmpp"),
+                actor,
+                object: { type: "credentials", password: "xmpp-pass" },
+            });
+            sc.socket.emit("credentials", {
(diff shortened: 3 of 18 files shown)


silverbucket:
Superseded — split into one PR per issue as requested:

- #1224 — detach a session when the platform rejects its credentials
- #1225 — return one uniform error for every refused session attach
- #1226 — scope stored credentials by platform as well as actor

Each is branched independently from `master` and reviewable on its own. Content is unchanged apart from the tests this branch added now using `it()` per AGENTS.md.

Branch `fix/session-hygiene-followups` left in place for reference; safe to delete.

PR #1225, opened 2026-08-16
fix(data-layer): return one uniform error for every refused session attach

silverbucket:
One of three followups split out of the #1131 review. Independent of the ID-format discussion, and independent of #1222.

## Problem

`credentials not found for X`, `invalid credentials for X` and `username already in use` were three distinguishable answers to an attach attempt, so a caller probing guessed actor ids learned *why* it was refused.

## Fix

All attach failures return a single `SESSION_SHARE_DENIED` message.

Error **classes** are unchanged, so behaviour that keys off them is preserved:
- same-IP anonymous reconnect still keys off `CredentialsNotShareableError`
- a missing-credentials failure is deliberately still a plain `Error`, so it stays *ineligible* for that reconnect path

Detailed messages are kept for the platform's own credential reads, where the caller already owns the credentials being described.

## Scope — this narrows disclosure, it does not close it

Accept-vs-reject is inherently observable: an attacker can still infer that *some* session holds a given actor id. Only the reason is now hidden. Fully closing it would mean rejecting unknown actors too, which isn't viable. Calling that out so the PR isn't read as more than it is.

## Behaviour change

The client-visible error for a refused share is now `unable to attach session for this actor`. `docs/client-guide.md`, `docs/configuration.md` and `packages/server/README.md` are updated.

## Tests

- all four rejection paths (no credentials stored, empty credential object, no secret present, and the existing anonymous case) return the identical message
- detailed messages survive for non-share reads

`bun test` 820 pass / 0 fail. `biome check`, `markdownlint` on changed docs, and build clean.

## Related

- #1222 — verify credentials match before attaching. Touches `credential-check.ts`; whichever merges second needs a trivial rebase.
- #1224, #1226 — the other two followups

### DIFF
--- a/docs/client-guide.md
+++ b/docs/client-guide.md
@@ -254,7 +254,7 @@ The rejected request returns an ActivityStream response with an `error`:
   "type": "connect",
   "actor": { "id": "mynick@irc.libera.chat", "type": "person", "name": "mynick" },
   "id": "1",
-  "error": "username already in use"
+  "error": "unable to attach session for this actor"
 }
 ```
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -246,8 +246,9 @@ When another socket is already attached to that actor:
 
 - Session-share validation runs in the data layer.
 - Credentials with a non-empty `password` are considered shareable.
-- Credentials without a non-empty `password` are not shareable and return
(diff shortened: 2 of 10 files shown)


silverbucket:
@coderabbitai code review

